PR #9, opened 2017-10-10
Enable Auth0 login

andyhd:
## What

* Add login with Auth0
* Add `ensureLoggedIn` middleware to protect views
* Add `login` and `logout` views in base app

## How to review

1. Add the following environment variables:

| Variable name  | Value  |
| ------------- | ----- |
| `AUTH0_DOMAIN`  | *Domain* from client settings page in Auth0  |
| `AUTH0_CLIENT_ID`  | *Client ID* from client settings page in Auth0  |
| `AUTH0_CLIENT_SECRET`  | *Client Secret* from client settings page in Auth0  |
| `AUTH0_CALLBACK_URL`  | `http://localhost:3000/callback`  |

2. Run the app
3. Click the `Login` button in the top right on the home page
4. Get prompted to login via Auth0
5. Click `Logout` link
6. Click `Apps` link and get redirected to login again


app/base/views.js
```diff
@@ -1,3 +1,28 @@
+var bole = require('bole');
```
xoen: don't think `bole` is used in this file.

app/config.js
```diff
@@ -16,9 +16,16 @@ config.log = {
 
 config.apps = [
   'base',
+  'apps',
   'users'
 ];
 
+config.session = {
+  secret: process.env.COOKIE_SECRET || 'shh-its-a-secret',
```
xoen: I would probably avoid a default in this case or even throw an exception in case this env. variable is missing (misconfiguration).

PS: This is just a minor observation so feel free to ignore if you reckon is not important, etc...

app/templates/home.html
```diff
@@ -3,7 +3,39 @@
...
 <a href="{{ url_for('users.list') }}">Users</a>
 {% endblock %}
 
-{% block right_column %}&nbsp;{% endblock %}
+{% block right_column %}
+{% if session.passport.user %}
+Hello, {{ session.passport.user.name }}
+<a href="{{ url_for('base.logout') }}">Logout</a>
+{% else %}
+<button id="login">Login</button>
```
xoen: Couldn't we potentially make this `login` link a link to `/login` and then drop the `login()` function here? Maybe I'm missing something but it seems like it's a duplication of what's happening in that view.

xoen: (Also if for whatever reason we need to change something login-related we'd then do only in that view)

xoen:
My knowledge of this stuff is limited but as far as I can tell it looks good 👍 (also, see comments)

app/templates/home.html
andyhd: This is meant as an example of different ways of invoking Auth0

app/config.js
andyhd: The secret MUST be set explicitly in the environment in deployment, but it's not very important in dev, so the default just makes life easier. All the defaults are (should be) tuned for running in development.

app/base/views.js
andyhd: You are right, I had some debug logging and removed it, but forgot to remove the `require`

app/templates/home.html
axemonkey: Yep.

PR #14, opened 2017-10-18
Use Auth0 ID token JWT to authenticate with API

andyhd:
## What

* Update api-client to authenticate with `JWT <token>` authorization header
* Added some template globals to make it more convenient to access current user, request and others

## How to review

1. Run the [control panel API](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/42)
2. Ensure you have the correct values in `.env`:

| name | value |
| ---- | ----- |
| `AUTH0_DOMAIN` | The `Domain` value from the dev `API client` in Auth0 |
| `AUTH0_CLIENT_ID` | The `Client ID` value |
| `AUTH0_CLIENT_SECRET` | The `Client Secret` value |
| `AUTH0_CALLBACK_URL` | `http://localhost:3000/callback` |

3. `npm start`
4. Login with Auth0
5. See user data retrieved from the API

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -32,9 +32,11 @@ passport.use(new Strategy({
     domain: process.env.AUTH0_DOMAIN,
     clientID: process.env.AUTH0_CLIENT_ID,
     clientSecret: process.env.AUTH0_CLIENT_SECRET,
-    callbackURL: process.env.AUTH0_CALLBACK_URL
+    callbackURL: process.env.AUTH0_CALLBACK_URL,
+    passReqToCallback: true
   },
-  function (issuer, audience, profile, cb) {
+  function (req, issuer, audience, profile, accessToken, refreshToken, params, cb) {
+    profile._json.id_token = params.id_token;
     return cb(null, profile._json);
   }
 ));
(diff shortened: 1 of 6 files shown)


app/templates/home.html
```diff
@@ -1,19 +1,19 @@
...
-  {% if session.passport.user %}
+  {% if req.user %}
   <a href="{{ url_for('base.home') }}" class="button button-secondary">Dashboard</a>
-  {% for key, value in session.passport.user %}
-  <p>{{ key }}: {{ value }}</p>
-  {% endfor %}
+  {% if error %}
+    <p>{{ error }}</p>
+  {% else %}
+    {{ user | dump | safe }}
```
axemonkey: Ahhhhh. Handy. 👌

app/index.js
```diff
@@ -32,9 +32,11 @@ passport.use(new Strategy({
     domain: process.env.AUTH0_DOMAIN,
```
axemonkey: Mind you... if we're changing `AUTH0` to `OIDC` in the CP app, shouldn't we do the same here?

andyhd: yes

axemonkey: 👍

PR #18, opened 2017-10-24
Fix login / home templates

axemonkey:
## What

We had login and "home" on the same template which were being context switched. This kicks a non-logged-in user to `/login` and then returns them to `/` when they authenticate.

## How to review

1. check out branch and visit root of site
2. observe that you are redirected to `/login`
3. log in with Auth0
4. observe that you now arrive at `/` and your user is dumped to page

### DIFF
--- a/app/base/views.js
+++ b/app/base/views.js
@@ -15,7 +15,7 @@ exports.home = function (req, res) {
       .catch(function (error) { render({'error': error}); });
 
   } else {
-    render({'user': {}});
+    res.redirect('/login');
   }
 };
 
--- a/app/base/views.js
+++ b/app/base/views.js
@@ -1,23 +1,22 @@
 var api = require('../../lib/api-client');
 var passport = require('passport');
+var ensureLoggedIn = require('connect-ensure-login').ensureLoggedIn;
 
 
-exports.home = function (req, res) {
+exports.home = [
+  ensureLoggedIn('/login'),
+  function (req, res) {
+    function render(context) {
+      res.render('home.html', context);
+    }
 
-  function render(context) {
-    res.render('home.html', context);
-  }
-
-  if (req.user) {
     api.authenticate(req.user.id_token);
     api.users.get(req.user.sub)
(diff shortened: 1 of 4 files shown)


app/templates/includes/proposition-links.html
```diff
@@ -1,6 +1,6 @@
-{% if session.passport.user %}
+{% if req.user %}
 <ul id="proposition-links">
-  <li><a href="{{ url_for('base.home') }}">Signed in as {{ session.passport.user.name }}</a></li>
+  <li><a href="{{ url_for('base.home') }}">Signed in as {{ user.name if user.name else user.email }}</a></li>
```
andyhd: I think these want to be `req.user.name` and `req.user.email`, because `user` may not always be passed to the template, but `req` is always available in templates (because [I added it to `app.locals`](https://github.com/ministryofjustice/analytics-platform-control-panel-frontend/blob/4f3a68fa1a24646587f7311c08ae86876aab410d/app/index.js#L55)).

Or, these could use `current_user` - which is also in `app.locals`. We could change that to `user` and then this code wouldn't have to change.

app/base/views.js
```diff
@@ -15,7 +15,7 @@ exports.home = function (req, res) {
       .catch(function (error) { render({'error': error}); });
 
   } else {
-    render({'user': {}});
+    res.redirect('/login');
```
andyhd: This could be done by adding `ensureLoggedIn('/login')` to the route, eg:
```javascript
var ensureLoggedIn = require('connect-ensure-login').ensureLoggedIn;

exports.home = [
  ensureLoggedIn('/login'),
  function (req, res) {

    function render(context) { ... }

    api.authenticate(...);
    api.users.get(req.user.sub)
      .then(...)
      ...
  }
];

app/templates/includes/proposition-links.html
axemonkey: Fine with changing this to `current_user` – if I'd thought to look in `app.locals` I would have used that anyway.

PR #22, opened 2017-10-26
Report errors to Sentry

andyhd:
## What

* Add dependency on the `raven` Sentry client
* Configure Sentry client with new `SENTRY_DSN` environment variable
* Send errors and uncaught promise rejections to Sentry
* Added `morgan` for HTTP request logging

## How to review

1. Set the `SENTRY_DSN` environment variable
2. Run the app
3. Browse to `/error` - a testing page
4. See an error message displayed
5. Browse to https://sentry.service.dsd.io/mojds/control-panel-frontend/
6. See the error recorded in Sentry

See [Trello #401](https://trello.com/c/ES3ohbQJ)

### DIFF
--- a/app/errors.js
+++ b/app/errors.js
@@ -1,7 +1,16 @@
+var log = require('bole')('error-handler');
+var sentry = require('raven');
 var url_for = require('./routes').url_for;
 
 
+process.on('unhandledRejection', function (error) {
+  log.error(error.message);
+  sentry.captureException(error);
+});
+
+
 module.exports = function (err, req, res, next) {
+  sentry.captureException(err);
 
   if (res.headersSent) {
     return next(err);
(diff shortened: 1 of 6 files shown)


app/errors.js
```diff
@@ -1,7 +1,16 @@
...
 
 
+process.on('unhandledRejection', function (error) {
+  log.error(error.message);
+  sentry.captureException(error);
+});
+
+
 module.exports = function (err, req, res, next) {
+  sentry.captureException(err);
```
xoen: How does this line work? Will this send the exception to Sentry straight away?

app/errors.js
```diff
@@ -1,7 +1,16 @@
+var log = require('bole')('error-handler');
+var sentry = require('raven');
 var url_for = require('./routes').url_for;
 
 
+process.on('unhandledRejection', function (error) {
```
xoen: What is this `unhandledRejection` event?

xoen: Oh, [it's happening when a promise's error is not handled](https://nodejs.org/api/process.html#process_event_unhandledrejection).

xoen:
Looks OK to me 👍

app/errors.js
andyhd: Yes, this is the global error handler middleware, so it sends the exception to Sentry when an error is passed up the request handler chain.

PR #25, opened 2017-10-27
Patch musl libc CVE

andyhd:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-15650


andyhd:
This is unnecessary, as the patch has been applied in the base node 6.11.5 docker image

PR #28, opened 2017-10-30
Add endpoint to create app

xoen:
## What

Added endpoint to create app by making request to API.
Also re-render the app form by passing the API errors so that we can give the user some meaningful feedback.

### Ticket

https://trello.com/c/rouWbSy1/439-create-app-on-post-request-2


xoen:
Looks like some tests are failing so having a look, also I'll try to write some tests for this.

PR #30, opened 2017-10-31
Superuser template for listing buckets

axemonkey:
## What

Template and route/view for listing the buckets as a superuser

## How to review

1. Log in
2. Visit `/buckets`
3. See any buckets in your API listed

### DIFF
--- a/app/buckets/views.js
+++ b/app/buckets/views.js
@@ -7,7 +7,7 @@ var log = bole('buckets-views');
 
 exports.list_buckets = [
 
-  ensureLoggedIn('/'),
+  ensureLoggedIn('/login'),
   function (req, res, next) {
 
     api.list_buckets()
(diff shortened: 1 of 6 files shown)


app/buckets/views.js
```diff
@@ -1,9 +1,25 @@
 var api = require('../api-client');
 var bole = require('bole');
+var ensureLoggedIn = require('connect-ensure-login').ensureLoggedIn;
 
 var log = bole('buckets-views');
 
 
+exports.list_buckets = [
+
+  ensureLoggedIn('/'),
```
andyhd: doesn't this want to be `ensureLoggedIn('/login')` now?

axemonkey: Yep, quite prolly. It's wrong in a couple of other places too, which is where I copypasta'd it from. I'll correct them all and update. 👍

PR #31, opened 2017-10-31
Add bucket via API on POST of add bucket form

andyhd:
## What

* Refactored routing setup to allow a route to specify which HTTP method to match
* Added a view and route to handle a POST request from the add bucket form
* Added an integration test for the view

## How to review

1. Run the app
2. Login
3. Click on the `Create new bucket` button
4. Enter a name and submit the form
5. See in S3 that a bucket with the specified name has been created
6. See the JSON representation of the bucket model displayed

See [Trello #442](https://trello.com/c/scu4u4ZD)

### DIFF
--- a/app/api-client.js
+++ b/app/api-client.js
@@ -9,7 +9,12 @@ module.exports = (function () {
   var token = 'invalid token';
 
   function api_request(options) {
-    return request(override_defaults(options));
+    try {
+      return request(override_defaults(options));
+
+    } catch (error) {
+      throw new Error(`API: ${options.method || 'GET'} ${options.endpoint} failed: ${error}`);
+    }
   }
 
 
@@ -30,7 +35,16 @@ module.exports = (function () {
 
 
   function endpoint_url(endpoint) {
-    return url.resolve(config.api.base_url, '/' + (endpoint + '/' || ''));
+
+    if (!endpoint) {
+      throw new Error('Missing endpoint');
+    }
+
+    if (!endpoint.endsWith('/')) {
+      endpoint += '/';
+    }
+
+    return url.resolve(config.api.base_url, endpoint);
   }
 
 
--- a/app/api-client.js
+++ b/app/api-client.js
@@ -30,7 +30,7 @@ module.exports = (function () {
 
 
   function endpoint_url(endpoint) {
-    return url.resolve(config.api.base_url, '/' + (endpoint || ''));
+    return url.resolve(config.api.base_url, '/' + (endpoint + '/' || ''));
   }
 
 
--- a/app/routes.js
+++ b/app/routes.js
@@ -86,11 +86,28 @@ function add_route(app_name) {
 
     log.debug(route.name + ': ' + route.pattern);
 
-    exports.router.all(route.pattern, route.view);
+    if (!route.method) {
+      route.method = 'get';
+    }
+
+    route.method = route.method.toLowerCase();
+
+    if (valid_method(route.method)) {
+      exports.router[route.method](route.pattern, route.view);
+
+    } else {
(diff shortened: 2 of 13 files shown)


app/api-client.js
```diff
@@ -30,6 +30,9 @@ module.exports = (function () {
 
 
   function endpoint_url(endpoint) {
+    if (endpoint && !endpoint.endsWith('/')) {
```
xoen: Some thoughts on this (don't have to change anything if you don't want)

We should probably always pass the endpoint in to `api_request()`, I wonder if here default the `endpoint` param to `/` would be defensive enough.

If you have a default and this check you may also get rid of the `(endpoint || '')` below, and maybe just use string interpolation, e.g.

```javascript
return url.resolve(`${config.api.base_url}/${endpoint}`)
```

BTW, I like that we check if ends with `/` better :)

andyhd: Yep, I haven't been writing ES6, but I will.

axemonkey: I guess I'm going to have to bite the bullet and start ES6ing at some point...

andyhd: @xoen in this case, your suggestion doesn't work, because `url.resolve` does the concatenation of the URL parts - there are two parameters, not one.

app/routes.js
```diff
@@ -86,12 +86,29 @@ function add_route(app_name) {
...
+    } else {
+      throw new Error('Invalid method "' + route.method.toUpperCase() + '" in route ' + route);
+    }
 
     return route;
   };
 }
 
+
+function valid_method(method) {
```
xoen: In Ruby there is this amazing thing that you can end methods with question marks, e.g. `valid_method?` and you then know it would return a "boolean"...

...in absense of that I would probably call this `is_valid_method()` to make explicit it's returning a boolean.
I know it doesn't seem like a huge difference, especially when you looking at the function implementation, but it can helps avoiding confusion.

app/templates/buckets/new.html
```diff
@@ -4,9 +4,13 @@
 
 {% block main_column %}
 
-<form action="/buckets/create" method="post">
+{% if error and not error.error %}
+<div class="error">{{ error.message }}</div>
```
xoen: I tend to indent these to make clear the `<div>` would be displayed if the `if` condition is true.
Again, doesn't look like a massive difference in this case but when you have more "complicated" templates can be very helpful.

(And yes, the rendered output would have an indentation which may seem odd but as indentation is ignore in HTML I would err on the side of clarity for us)

app/routes.js
andyhd: Yeah, I'll change it. I personally don't like the flow of `if (is_valid_method(method)) {`, it sounds better as `if (valid_method(method))` or `if (is_valid(method))` - but the latter removes too much information from the function name.
Mind you, there's no guarantee that a function named `is_valid_method` will return a boolean, so naming conventions like that just feel like a step towards Hungarian notation (eg: `boolIsValidMethod`).

app/templates/buckets/new.html
andyhd: Agreed

app/routes.js
andyhd: Just reminded myself that I'm naming things with snake_case when they should really be camelCase in Javascript 😕

app/api-client.js
xoen: Mainly was a point around testing if the `endpoint` was present and use of ES6 concatenation, etc...but you got the idea 🖖

app/routes.js
xoen: Should they :trollface: ?

I know what you mean, I more used and prefer snake_case, and also when using different style in backend/frontend involves more work to convert cases it just becomes madness.

PR #32, opened 2017-11-01
List view for apps

axemonkey:
## What

List apps so that superadmin can see them all and manage them

## How to review

1. Run application and log in
2. Have some apps in your API DB
3. Visit `/apps`
4. See your apps listed with links to their detail pages and repos, and (non-working) management buttons

## Looky likey

screen shot 2017-11-01 at 12 24 31

### DIFF
--- a/app/templates/apps/list.html
+++ b/app/templates/apps/list.html
@@ -1,13 +1,35 @@
-{% extends "layouts/two-column.html" %}
+{% extends "layouts/one-column.html" %}
 
 {% set page_title = "Apps" %}
 
 {% block main_column %}
 
-<ul>
-  {% for app in apps.results %}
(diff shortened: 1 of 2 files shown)


app/templates/apps/list.html
```diff
@@ -1,13 +1,35 @@
...
+      <th>Description</th>
+      <th>Repo</th>
+      <td>&nbsp;</td>
+    </tr>
+  </thead>
+  <tbody>
+    {% for app in apps.results %}
+    <tr>
+      <td>{{ app.id }}</td>
+      <td><a href="/apps/{{ app.id }}">{{ app.name }}</a></td>
```
andyhd: This href should be `{{ url_for('apps.details', {id: app.id}) }}`

axemonkey: Ah, nice.

PR #33, opened 2017-11-01
Request user data from API on authentication

andyhd:
## What

* Make a request to the API to fetch the user that has just authenticated, to give the API a chance to create the user in the database if it does not exist.

## How to review

No obvious change, but there is a new passing test

See [Trello #451](https://trello.com/c/lBz5zvth)

### DIFF
--- a/app/base/views.js
+++ b/app/base/views.js
@@ -27,10 +27,14 @@ exports.error_test = function (req, res, next) {
 
 exports.auth_callback = [
   passport.authenticate('auth0-oidc'),
-  function (req, res) {
+  function (req, res, next) {
     raven.setContext({user: req.user});
     api.set_token(req.user.id_token);
-    res.redirect(req.session.returnTo || '/');
+    api.users.get(req.user.sub)
+      .then((user) => {
+        res.redirect(req.session.returnTo || '/');
+      })
+      .catch(next);
   }
 ];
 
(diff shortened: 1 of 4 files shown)


app/base/views.js
```diff
@@ -27,10 +27,14 @@ exports.error_test = function (req, res, next) {
 
 exports.auth_callback = [
   passport.authenticate('auth0-oidc'),
-  function (req, res) {
+  function (req, res, next) {
     raven.setContext({user: req.user});
     api.set_token(req.user.id_token);
-    res.redirect(req.session.returnTo || '/');
+    api.users.get(req.user.sub)
```
xoen: So this will depend on James' work here: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/57

app/index.js
```diff
@@ -33,18 +33,12 @@ app.use(require('cookie-parser')());
...
-passport.use(new Strategy({
-    domain: process.env.AUTH0_DOMAIN,
-    clientID: process.env.AUTH0_CLIENT_ID,
-    clientSecret: process.env.AUTH0_CLIENT_SECRET,
-    callbackURL: process.env.AUTH0_CALLBACK_URL,
-    passReqToCallback: true
-  },
-  function (req, issuer, audience, profile, accessToken, refreshToken, params, cb) {
+passport.use(new Strategy(
+  config.auth0,
```
xoen: nice 👍

app/base/views.js
andyhd: Yes. It doesn't depend on him implementing it, because we're not doing anything with the response, but he is implementing the user creation in the API.

PR #34, opened 2017-11-01
Added endpoint and (placeholder!) view for editing an app

xoen:
### What

Node endpoint `GET /apps/:id/edit` which:
* Make requests to CP-API to get app with given ID, list of S3 buckets and list of users (concurrently)
* Renders `apps/edit.html` passing this data in the `app`, `buckets` and `users` variables

(Currently the `apps/edit.html` template is just a placeholder which shows this information, not the proper form, etc...)

### NOTE
These three requests to get this data can be performed independently so we make them concurrently. This means that the total time would not be the sum of these requests' times but just the time of the slower request.

### Ticket
https://trello.com/c/Das3TYNG/444-fetch-lists-of-buckets-and-users-for-grant-access-to-bucket-form-3

### TODO
- [X] Make it work
- [X] Make requests concurrently
- [ ] Write tests


app/apps/views.js
```diff
@@ -148,3 +151,28 @@ exports.app_details = function (req, res) {
...
+
+exports.app_edit = [
+  ensureLoggedIn('/login'),
+  function(req, res, next) {
+    // This is making three requests cuncurrently to get the app details,
+    // the list of bucket and the list of users.
+    // If all of them succeed it then renders `apps/edit.html`
+    app_request = api.get_app(req.params.id);
+    buckets_request = api.list_buckets();
+    users_request = api.list_users();
```
andyhd: These need `var` or `const`

app/apps/views.js
```diff
@@ -148,3 +151,28 @@ exports.app_details = function (req, res) {
...
+
+    Bluebird
+      .all([app_request, buckets_request, users_request])
```
andyhd: You might be able to [do this with native Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all)

app/apps/views.js
xoen: Ops, well spotted :)

We could probably start to [use ES6 `let` in place of `var`](https://softwareengineering.stackexchange.com/q/274342) - `let` has block scope, `var` function scope, conclusion was "Adopt let and const. Stop using var".

app/apps/views.js
xoen: I tried yesterday with no success, I want to write some tests and then I could play with this again and see if I can make it work with native Promises.

test/test-app-edit.js
```diff
@@ -0,0 +1,72 @@
+"use strict";
+
+const chai = require('chai');
+const should = chai.should();
```
andyhd: Please use `chai.assert` for consistency with other tests.

test/test-app-edit.js
```diff
@@ -0,0 +1,72 @@
...
+const chaiHttp = require('chai-http');
+chai.use(chaiHttp);
```
andyhd: `chaiHttp` variable is used only once, so just inline it.

app/apps/views.js
```diff
@@ -1,6 +1,9 @@
+var ensureLoggedIn = require('connect-ensure-login').ensureLoggedIn;
+var Bluebird = require('bluebird');
```
andyhd: This is not necessary

app/apps/views.js
```diff
@@ -148,3 +151,28 @@ exports.app_details = function (req, res) {
...
+    // This is making three requests cuncurrently to get the app details,
+    // the list of bucket and the list of users.
+    // If all of them succeed it then renders `apps/edit.html`
+    let app_request = api.get_app(req.params.id);
+    let buckets_request = api.list_buckets();
+    let users_request = api.list_users();
+
+    Bluebird
+      .all([app_request, buckets_request, users_request])
+      .spread(function(app_response, buckets_response, users_response) {
```
andyhd: Bluebird provides a non-standard Promises implementation. This can be achieved with the standard Promises API:
```javascript
Promise
  .all([app_request, buckets_request, users_request])
  .then((responses) => {
    let [app_response, buckets_response, users_response] = responses;
    ...
```

app/apps/views.js
```diff
@@ -148,3 +151,28 @@ exports.app_details = function (req, res) {
...
   });
 };
+
+
+exports.app_edit = [
+  ensureLoggedIn('/login'),
+  function(req, res, next) {
+    // This is making three requests cuncurrently to get the app details,
+    // the list of bucket and the list of users.
+    // If all of them succeed it then renders `apps/edit.html`
```
andyhd: This comment adds no information

package.json
```diff
@@ -19,6 +19,7 @@
   },
   "homepage": "https://github.com/ministryofjustice/analytics-platform-control-panel-frontend#readme",
   "dependencies": {
+    "bluebird": "^3.5.1",
```
andyhd: Not necessary

package.json
```diff
@@ -45,6 +46,7 @@
   "devDependencies": {
     "bistre": "1.0.1",
     "chai": "^4.1.2",
+    "chai-http": "^3.0.0",
```
andyhd: Not necessary

test/test-app-edit.js
andyhd: But also, it's not necessary

test/test-app-edit.js
```diff
@@ -0,0 +1,72 @@
...
+      // api.list_buckets().then(
+      //   (buckets) => {
+      //     console.log("BUCKETS =>", JSON.stringify(buckets))
+      //   }
+      // )
+      // api.list_users().then(
+      //   (users) => {
+      //     console.log("USERS =>", JSON.stringify(users))
+      //   }
+      // )
```
andyhd: Don't commit commented-out code

test/test-app-edit.js
```diff
@@ -0,0 +1,72 @@
...
+      // Make request to node endpoint
+      // SEE: https://scotch.io/tutorials/test-a-node-restful-api-with-mocha-and-chai
+      chai
+        .request(server)
+        .get(`/apps/${APP.id}`)
+        .end((err, res) => {
+          res.should.have.status(200);
+          res.body.app.should.be.eql(APP);
+          res.body.buckets.should.be.eql(BUCKETS);
+          res.body.users.should.be.eql(USERS);
```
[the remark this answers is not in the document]
andyhd: Using `chaiHttp` loads in the app, middleware and other dependencies unnecessarily. We don't need to load them to test the `app_edit` request handler.

andyhd:
I've fixed your tests, but I've submitted a PR on this PR instead of committing them directly to this branch.

xoen:
Superseded by other PR based on this one: https://github.com/ministryofjustice/analytics-platform-control-panel-frontend/pull/38. Closing this.

PR #37, opened 2017-11-06
Create user template for wiring up

axemonkey:
## What

Template for getting user details on user creation so that the user can verify them, and check the email address they want to use.

## How to review

1. Run app and log in
2. Visit `/users/new`
3. See the three fields


andyhd:
Merged with @axemonkey's permission

PR #42, opened 2017-11-07
Show a friendly message to non superusers

andyhd:
## What

* Catch API 403 error on login and set flag on user session to show a friendly error message instead of a 500 error page

## How to review

1. Set your user attribute `is_superuser` to `False` in your database

    ```sh
    docker-compose exec api python3 manage.py shell
    ```
    then
    ```python
    from control_panel_api.models import User
    me = User.objects.get(username='my username')
    me.is_superuser = False
    me.save()
    exit()
    ```
2. Login to the frontend
3. See a friendly error message

### DIFF
--- a/app/templates/home.html
+++ b/app/templates/home.html
@@ -4,8 +4,10 @@
 
 {% block main_column %}
 
-  {% if error %}
-    <p>{{ error }}</p>
+  {% if not current_user.is_superuser %}
+    <h2 class="heading-medium">No access yet</h2>
+    <p>Only superadmins have access to the control panel at the moment.</p>
+    <p>Ask a superadmin to grant you access.</p>
   {% else %}
     <h2 class="heading-medium">Admin</h2>
     <ul class="list list-bullet">
--- a/app/base/views.js
+++ b/app/base/views.js
@@ -1,19 +1,14 @@
-var api = require('../api-client');
-var passport = require('passport');
-var ensureLoggedIn = require('connect-ensure-login').ensureLoggedIn;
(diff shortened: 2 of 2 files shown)


app/templates/home.html
```diff
@@ -4,8 +4,10 @@
 
 {% block main_column %}
 
-  {% if error %}
-    <p>{{ error }}</p>
+  {% if not current_user.is_superuser %}
+  <h2 class="heading-medium">No access yet</h2>
```
xoen: As discussed in another PR, I always prefer to indent in the templates to make them more readable and avoid confusion/bugs.

app/templates/home.html
```diff
@@ -4,8 +4,10 @@
 
 {% block main_column %}
 
-  {% if error %}
-    <p>{{ error }}</p>
+  {% if not current_user.is_superuser %}
```
xoen: Not sure I'm keen on this approach. It works but it's a bit strange that the UI determine if the user is a super user, and also we set this flag to show a permission error message.

I'm thinking, maybe we can set something like `req.errors` or something like that.

app/templates/home.html
andyhd: Yep, I knew you would say that - I will change

app/templates/home.html
andyhd: This is just a stop-gap solution while we wait for proper authorization in the API. Then we'll get the permissions from the API and check those instead.

xoen: 👍

PR #43, opened 2017-11-08
Add `GET /users/:id/edit` endpoint

xoen:
### What

This endpoint renders `users.edit.html` passing the following context:

```json
{
    "user": {},
    "apps_options": [],
    "buckets_options": []
}
```

Where the `(apps|buckets)_options` is the list of apps/buckets not already associated with the user.

### Ticket

https://trello.com/c/nOIbztRT/448-fetch-lists-of-buckets-and-apps-for-grant-access-to-user-form-1

### Checklist
- [x] Filter out apps/buckets already associated with user

### DIFF
--- a/app/templates/users/edit.html
+++ b/app/templates/users/edit.html
@@ -1,0 +1,13 @@
+{% extends "layouts/one-column.html" %}
+
+{% set page_title = "User: " + user.username %}
+
+{% block main_column %}
+
+<pre>user = {{ user | dump(4) | safe }}</pre>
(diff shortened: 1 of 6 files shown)


xoen:
@axemonkey the `user` coming from the API in this case contains the information to filter out `apps`/`buckets` already associated so I can try to change so that `apps`/`buckets` passed to the view contains only stuff *not* associated with the user already.

My understanding is that it's what you'd prefer, thoughts?

PR #45, opened 2017-11-10
Use Redis for session storage

andyhd:
## What

* Use Redis for session storage to persist sessions across server restarts - meaning you don't have to repeatedly log in after changing a file in development
* Refactored app initialization to a modular style

## How to review

1. Run the app with `docker-compose up`
2. Login
3. Restart the server (by changing a file, for example)
4. Visit a page which requires login
5. See you don't have to login again


Dockerfile
```diff
@@ -14,6 +14,8 @@ ADD node_modules node_modules/
 RUN npm run-script collect-static
 ADD static static/
 
+RUN mkdir -m 777 sessions
```
xoen: Just started to have a look at this so I don't know very well who will access `sessions` but do we need this level of permissions?

Who will read or write from this directory? If it's just the current user [`600 / (only owner can read/write)`](http://www.filepermissions.com/file-permission/600) may be more than plenty.

app/assets.js
```diff
@@ -1,101 +1,102 @@
...
-var concatfiles = require('concat-files');
+const config = require('./config');
+const log = require('bole')('assets');
 
 
-exports.compile_sass = function () {
+exports.compile_sass = () => {
 
-  config.sass.sources.forEach(function (options) {
+  config.sass.sources.forEach((options) => {
```
xoen: You already know this because you did it in a different part of this PR, but you could omit the parenthesis around options as the arrow function has only one argument.

I still have to make up my mind about this "special" case as parenthesis are mandatories if function has 0 or 2+ arguments - but I admit that it's nice to not have the double parenthesis in cases like this or when function is very small so I'd probably remove them if optional.

app/assets.js
```diff
@@ -1,101 +1,102 @@
-var config = require('./config');
-var log = require('bole')('assets');
-var path = require('path');
+const Bluebird = require('bluebird');
+const fs = Bluebird.promisifyAll(require('fs'));
+const glob = Bluebird.promisify(require('glob').glob);
```
xoen: I see that `Bluebird.promisifyAll()` adds the `Async` suffix. I wonder if for consistency and for the sake of being explicit we should add this `Async` suffix to the name of all the "promisified" - just a though.

app/base/views.js
```diff
@@ -56,5 +57,8 @@ exports.login = function (req, res) {
 exports.logout = function (req, res) {
   req.logout();
   api.unset_token();
-  res.redirect('/');
+  req.session.destroy(err => {
```
xoen: If `err` is not used in the handler, you may want to rename this to `_` to make it clear it's an unused argument.

app/factory.js
```diff
@@ -0,0 +1,37 @@
+const express = require('express');
```
xoen: May be the case of renaming the `factory` => `app-factory` to be more accurate.

app/assets.js
xoen: The more I think about it and look at this PR the more I think omit the parenthesis when the arrow function only take only one argument makes it more readable (and we're already doing this).

app/config.js
```diff
@@ -1,17 +1,18 @@
-var join = require('path').join;
-var config = module.exports;
-var PROD = process.env.ENV === 'prod';
+const join = require('path').join;
 
-var node_modules = join(__dirname, '../node_modules');
 
+const config = module.exports;
```
xoen: I don't know if I'm missing something but I'm wondering if all these `config.something = {}` could be replaced with a single big:

```javascript
module.exports = {
  express: {
    // ...
  },
  middleware = [
    // ...
  ],
  // etc...
}
```

app/factory.js
```diff
@@ -0,0 +1,37 @@
...
+  const nunjucks = require('nunjucks');
+  nunjucks.configure(join(__dirname, 'templates'), {
+    autoescape: true,
+    express: app
+  });
+}
+
+
+function init_middleware(app, conf) {
+  const middleware = require('./middleware');
```
xoen: I would rename this module `middleware` to `middlewares` as this is the "map" with the middleware**s**, I'd also rename `contig.middleware` to the plural.

You may want to rename other variables to make the distinction between strings and functions a bit more explicit (I didn't realise about this distinction when I started to have a look at this code):

```javascript
const middlewareNames = require('./middleware');
middlewareNames.forEach(name => {
  const middleware = config.middlewares[name];
  app.use(middleware(app, config));
});
```

app/index.js
```diff
@@ -1,72 +1,10 @@
...
-var join = require('path').join;
-var nunjucks = require('nunjucks');
-var passport = require('passport');
-var raven = require('raven');
-var Strategy = require('passport-auth0-openidconnect').Strategy;
-var assets = require('./assets');
+const bole = require('bole');
 
+const config = require('./config');
+const create_app = require('./factory');
```
xoen: As discussed offline you may want to rename this module to `create_app` OR export using `module.exports.create_app = () => {}` and then here change this to `const create_app = require('./app_factory').create_app;`.

BTW: I like how you moved things around, it makes things a bit less daunting and easier to follow 👍

app/middleware.js
```diff
@@ -0,0 +1,122 @@
+const log = require('bole')('middleware');
```
xoen: I'd rename this module to `middlewares`.

Dockerfile
andyhd: I meant to remove this, actually - we're not storing sessions in the filesystem

app/routes.js
```diff
@@ -1,125 +1,84 @@
...
+const join = require('path').join;
 
 
 exports.router = new express.Router();
 
-var routes = [].concat.apply([], config.apps.map(add_app_routes));
+const routes = merge(config.apps.map(app_name => {
 
-exports.url_for = function (view_name, args) {
+  return merge(require(join(__dirname, app_name, 'routes')).map(route => {
```
xoen: Maybe it's because I'm not much familiar with this part of the code but I wonder if this `map` within a `map` can be a bit confusing.

app/assets.js
andyhd: I'm also undecided on the parenthesis. Double parentheses are ugly, but removing them contributes to the same issue as omitting braces on a single line `if` block. I think we should always use the parentheses even with only 1 argument.

app/routes.js
xoen: Even if not strictly necessary probably split in a couple of statements and use variables name to document.

app/assets.js
andyhd: No - that would lead to ever longer function names. I can't help what Bluebird does to wrapped functions, but I don't agree with it.

app/base/views.js
andyhd: I have done that sometimes, but it's a Python (maybe Ruby) idiom and there's a (unfortunately named) popular Javascript library called [Underscore](http://underscorejs.org/) which uses the `_` character and muddies the waters.

app/factory.js
andyhd: As discussed, I think I'll rename the `factory.js` to `create_app.js` or similar.

app/config.js
andyhd: But how would that improve things?

andyhd: I actually prefer this style (not my idea, I got it from someone else), because you always reference config variables in the same style `config.foo.bar`, even in the definition.

app/factory.js
andyhd: The word `middleware` is already plural.

In your snippet here, you're confusing the actual middleware (the objects in the middleware module) with the config array, so I take your point.

I'll try to name these better.

app/base/views.js
xoen: Also a Go idiom so almost a standard de-facto :D - I only said it as we have an `_` already in one of the files, but I imagine if underscore.js is also used it could cause a bit of confusion.

app/middleware.js
andyhd: This is the logger name, not a module. Log messages with be prefixed with (for example) `debug middleware: some log message`

app/routes.js
andyhd: Yep, as discussed, it's a bit dense. I will try to simplify it.

app/config.js
xoen: I suppose in mind the reason why I was saying it was because I thought about it as a big JSON config file - but you're right change doesn't probably give us much 👍 (and the way it is clear to me)

app/assets.js
```diff
@@ -86,15 +86,15 @@ function compile_all(files, base_options) {
...
+  return (result) => {
     fs.writeFileAsync(filename, result.css);
     return result;
   };
 }
 
 
 function write_sourcemap(options) {
-  return result => {
+  return (result) => {
```
xoen: 👍

app/routes.js
```diff
@@ -5,23 +5,22 @@ const join = require('path').join;
 
 exports.router = new express.Router();
 
-const routes = merge(config.apps.map(app_name => {
+const routes = {};
```
xoen: I'm finding this version of the logic much more readable than the previous version with all the merges 👍

xoen:
I like the changes and I think `app/routes.js` is much better now 👍

PR #47, opened 2017-11-14
Add endpoint `POST /users3buckets`

xoen:
### What

- Added endpoint `POST /users3buckets` to grant user access to a bucket.
- Merged `GET /buckets/:id/edit` into `GET /buckets/:id`
- Added placeholder `bucket-details` template with:
  - list of users already having access to the bucket (and access level)
  - form to grant access to other users

### More details
~~This endpoint is very similar to the `POST /apps/:id/connect_bucket` endpoint.
The idea is to give more natural names to these endpoints from now on.~~
Similar to this endpoint there will be others like:
- `POST /users3buckets/:id` (to update access)
- `POST /users3buckets/:id/destroy` (to revoke access)
- etc...

**NOTE**: [Didn't change the existing "connect_bucket" endpoint](https://trello.com/c/nvOCtrlY/472-1-rename-appsconnectbucket-appsgrantbucketaccess) to avoid conflicts as Clive was working on it - but I discussed this new naming convention with him already.

**NOTE**: The template is mostly copied from existing `app-details` template and it's only a temporary placeholder - @axemonkey will have a look at it and amend as required.

### Ticket
https://trello.com/c/OxRsnwPB/443-grant-user-access-to-bucket-on-post-request-2

### TODO

- [x] Add some tests


app/api-client.js
```diff
@@ -103,6 +104,7 @@ module.exports = (function () {
     return api_request({method: 'POST', endpoint: 'apps/', resource: app});
   };
 
+  // TODO: Rename `connect_bucket_to_app` => `grant_app_access_to_bucket`
```
andyhd: Don't leave TODOs in the code. Either do them, or add a ticket to Trello.

app/api-client.js
```diff
@@ -134,10 +137,15 @@ module.exports = (function () {
     return api_request({method: 'POST', endpoint: 's3buckets', resource: bucket});
   };
 
+  api.grant_user_access_to_bucket = function (users3bucket) {
```
andyhd: This function signature is confusing, it should either be:
```javascript
api.grant_user_access_to_bucket = function (user, bucket) {
```
or
```javascript
api.create_users3bucket = function (users3bucket) {
```
Also, the implementation detail that the data source is an S3 bucket should not be revealed in these methods, because the frontend should not need to know about S3 or buckets.

app/users/routes.js
```diff
@@ -2,10 +2,9 @@ var views = require('./views');
 
 
 module.exports = [
-
   {name: 'new', pattern: '/users/new', view: views.new_user},
   {name: 'list', pattern: '/users', view: views.list_users},
   {name: 'details', pattern: '/users/:id', view: views.user_details},
   {name: 'edit', pattern: '/users/:id/edit', view: views.user_edit},
-
+  // {name: 'grant_bucket_access', method: 'POST', pattern: '/users/:id/grant_bucket_access', view: views.grant_bucket_access},
```
andyhd: Remove commented out code. Is this supposed to be commented out?

app/api-client.js
xoen: 😭

xoen: Copy that: https://trello.com/c/nvOCtrlY/472-1-rename-appsconnectbucket-appsgrantbucketaccess 

FYI: @axemonkey I added that ticket, I could rename it myself later or you could do it if you feel like it as part of your current set of changes.

app/api-client.js
xoen: I will rename the *api* function to `api.create_users3bucket()` to make it more explicit and I may play with some ideas to make the API a bit nicer then 👍

app/users/routes.js
xoen: My idea was to leave an hint to the future person who was going to implement this feature.
Basically that person had simply to uncomment this and add a template, etc...without the risk of picking a name for this endpoint which was different.

app/buckets/routes.js
```diff
@@ -7,6 +7,5 @@ module.exports = [
   {name: 'new', pattern: '/buckets/new', view: views.new_bucket},
   {name: 'create', pattern: '/buckets/new', view: views.create_bucket, method: 'POST'},
   {name: 'details', pattern: '/buckets/:id', view: views.bucket_details},
-  {name: 'edit', pattern: '/buckets/:id/edit', view: views.bucket_edit},
-
+  {name: 'grant_user_access', method: 'POST', pattern: '/buckets/:id/grant_user_access', view: views.grant_user_access},
```
andyhd: This URL is a strange hybrid of REST and RPC. RESTful URLs should not really contain verbs (eg: grant), because the verb is the HTTP method (eg: POST).

A more RESTful URL might be `/buckets/:bucket_id/access` where you might POST a payload like `{user: user_id, access_level: 'readwrite'}`.

andyhd: Whether the frontend should have RESTful URLs, I'm not really sure.

app/users/routes.js
andyhd: ok

app/users3buckets/handlers.js
```diff
@@ -0,0 +1,25 @@
+const { ensureLoggedIn } = require('connect-ensure-login');
+
+const api = require('../api-client');
+const routes = require('../routes');
+
+
+exports.create = [
+  ensureLoggedIn('/login'),
+  (req, res, next) => {
+    const { user_id, bucket_id } = req.body;
```
andyhd: 👍

PR #49, opened 2017-11-14
Show actual buckets on create app page

axemonkey:
## What

Show actual buckets from API in dropdown instead of dummy data. Was intended as part of a "connect bucket during API creation" PR, but parking that for now.

## How to review

1. check out, run app, log in
2. go to `/apps/new`
3. scroll down and select "Connect an existing data source"
4. check the dropdown which appears – it should contain actual bucket data from your API rather than `dev-dummy-bucket-[1-5]`

### DIFF
--- a/app/apps/views.js
+++ b/app/apps/views.js
@@ -7,12 +7,8 @@ var routes = require('../routes');
 exports.new_app = [
   ensureLoggedIn('/login'),
   function (req, res, next) {
-    const buckets_request = api.list_buckets();
-
-    Promise
-      .all([buckets_request])
-      .then(function(response) {
-        const [buckets_response] = response;
+    api.list_buckets()
+      .then((buckets_response) => {
         const template_args = {
           prefix: process.env.ENV + '-',
           buckets: buckets_response.results,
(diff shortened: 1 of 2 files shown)


app/apps/views.js
```diff
@@ -5,37 +5,21 @@ var routes = require('../routes');
...
-        },
-        {
-          id: 5,
-          name: 'dev-dummy-bucket5'
-        },
-      ]
-    });
-
+    Promise
+      .all([buckets_request])
```
xoen: As you make only a request to get the list of buckets (`GET /s3buckets`), you can get rid of  ` Promise.all` and replace the deconstruction thing with something like:

```javascript
api.list_buckets()
     .then((buckets_response) => {
        const template_args = {
          prefix: process.env.ENV + '-',
          buckets: buckets_response.results,
        };
        res.render('apps/new.html', template_args);
      });
      .catch(next);
     })
```

(code above should work but it's obviously untested)

xoen:
See comment but as far as I can tell looks good to me 👍

PR #51, opened 2017-11-14
Log Redis errors and use password

andyhd:
## What

* Configure Redis to log errors to console
* Use a password for Redis to match deployment

## How to review

1. Run `docker-compose up`
2. See Redis error log message `WARNING: The REDIS_PASSWORD variable is not set. Defaulting to a blank string.` and possibly other error messages.
3. Note that you cannot progress past the login screen.
3. Add a `REDIS_PASSWORD` env var to your `.env` file
4. Re-run `docker-compose up`
5. See no error message from Redis

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -108,7 +108,9 @@ config.session = {
 
 config.session_store = {
   host: 'redis',
-  port: 6379
+  port: 6379,
+  logErrors: true,
+  pass: process.env.REDIS_PASSWORD
 };
 
 config.static = {
--- a/app/config.js
+++ b/app/config.js
@@ -107,7 +107,7 @@ config.session = {
 };
 
 config.session_store = {
-  host: 'redis',
+  host: process.env.REDIS_HOST || 'redis',
   port: 6379,
   logErrors: true,
   pass: process.env.REDIS_PASSWORD
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,5 +47,8 @@ services:
       POSTGRES_DB: "controlpanel"
   redis:
     image: "redis:4.0.2-alpine"
+    env_file:
+      - .env
     ports:
       - "6379:6379"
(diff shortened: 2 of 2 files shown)


andyhd:
That is the first error message, yes. You didn't get the one I mentioned eventually?

axemonkey:
Nope.

```
redis_1     | 1:C 15 Nov 14:04:08.841 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
redis_1     | 1:C 15 Nov 14:04:08.843 # Redis version=4.0.2, bits=64, commit=00000000, modified=0, pid=1, just started
redis_1     | 1:C 15 Nov 14:04:08.843 # Configuration loaded
redis_1     | 1:M 15 Nov 14:04:08.850 * Running mode=standalone, port=6379.
redis_1     | 1:M 15 Nov 14:04:08.850 # WARNING: The TCP backlog setting of 511 cannot be enforced because /proc/sys/net/core/somaxconn is set to the lower value of 128.
redis_1     | 1:M 15 Nov 14:04:08.850 # Server initialized
redis_1     | 1:M 15 Nov 14:04:08.850 # WARNING you have Transparent Huge Pages (THP) support enabled in your kernel. This will create latency and memory usage issues with Redis. To fix this issue run the command 'echo never > /sys/kernel/mm/transparent_hugepage/enabled' as root, and add it to your /etc/rc.local in order to retain the setting after a reboot. Redis must be restarted after THP is disabled.
redis_1     | 1:M 15 Nov 14:04:08.850 * Ready to accept connections
```

Those are the subsequent redis lines that docker is showing me.

andyhd:
Yes, but the point is that it is showing you an error message.

Does it still not work if you add a `REDIS_PASSWORD` setting to your .env file and run `docker-compose up`?

axemonkey:
That does work. So without one, it gives the message I posted before and just fails silently reloading the login page. With a PW it works.

PR #52, opened 2017-11-15
Add eslint and config

andyhd:
## What

* Add dependency on eslint
* Add dependency on airbnb eslint config
* Add .eslintrc.js with one rule override - don't complain about snake_case variable names
* Add npm script to run eslint

## How to review

1. `npm install` to install new dependencies
2. `npm run lint` to run eslint
3. See loads of warnings and errors


xoen:
Looks good to me...I'm afraid to run `nom run lint` :P

PR #56, opened 2017-11-16
Add endpoints to update/delete apps3buckets

xoen:
### What
Endpoints:
- `POST /apps3buckets/:id` to update access level
- `POST /apps3buckets/:id/delete` to revoke access

I also updated the template to replace the anchors with forms
(*need* styling from Clive).

Manually tested locally, I'll add some automated tests.

### Why `POST`?
It's only safe to use `POST` and `GET` as form methods in the browsers.
As these requests have a side effects we need to use `POST`.

### Why to pass `redirect_to`?
As this requests could be made from both the `apps.detail` page or the `buckets.details` page we need to know where to redirect the user on success.
Instead of guessing I'm being explicit and just passing this in the form (where I have this information).

### Ticket

https://trello.com/c/yYFb0Tax/475-2-add-endpoint-to-update-app-access-level-to-bucket

- [x] Add tests

### DIFF
--- a/test/test-apps3buckets-create.js
+++ b/test/test-apps3buckets-create.js
@@ -1,41 +1,34 @@
-"use strict";
-
-const assert = require('chai').assert;
+const { assert } = require('chai');
 const nock = require('nock');
 
 const config = require('../app/config');
 const handlers = require('../app/apps3buckets/handlers');
-const url_for = require('../app/routes').url_for;
+const { url_for } = require('../app/routes');
 
 
 describe('Edit bucket form', () => {
-
   describe('when granting access to user', () => {
-
     it('make request to API', () => {
       const bucket_id = 42;
       const app_id = 123;
 
       // Mock `POST /apps3buckets` request
       const post_apps3buckets = nock(config.api.base_url)
-        .post(`/apps3buckets/`, {
+        .post('/apps3buckets/', {
           app: app_id,
           s3bucket: bucket_id,
-          access_level: "readonly",
+          access_level: 'readonly',
         })
         .reply(201);
 
-      let request = new Promise((resolve, reject) => {
-        let req = {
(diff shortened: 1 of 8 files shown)


xoen:
@axemonkey I'm going to write some automated tests for these handlers but you can start to have a look if you'd like. I don't imagine the code changing much.

xoen:
It appears Clive already have classes for the buttons, superstar!! 🌟🌟🌟

test/test-apps3buckets-create.js
```diff
@@ -1,55 +1,46 @@
...
+          },
         };
 
         handlers.create[1](req, res, reject);
       });
 
       return request
         .then((redirect_url) => {
-          assert(post_apps3buckets.isDone(), `Make POST request to /apps3buckets (API)`);
+          assert(post_apps3buckets.isDone(), 'Make POST request to /apps3buckets (API)');
```
andyhd: The string parameter is the error message shown if the assertion fails. It should explain why the assertion failed, eg: "Did not make POST request to API endpoint /apps3buckets"

xoen: 👍

test/test-apps3buckets-delete.js
```diff
@@ -0,0 +1,40 @@
...
+const handlers = require('../app/apps3buckets/handlers');
+
+
+describe('Edit bucket form', () => {
+  describe('when revoking access to user', () => {
+    it('make request to API', () => {
+      const apps3bucket_id = 42;
+      const redirect_to = 'apps/123';
+
+      // Mock `DELETE /apps3buckets/:id` request
```
andyhd: This comment doesn't add any information. Comments are to explain _why_ the code is doing something, not _what_ it is doing.

PR #57, opened 2017-11-16
Delete app endpoint and wired up buttons

axemonkey:
## What

Allows superuser to delete apps in the UI

## How to review

1. check out, run, log in
2. create an app
3. visit `/apps`, then either
4. click "Delete app" in the table, or
5. click through to the app page and click "Delete app" at the bottom of the page
6. see that the app has been deleted

### DIFF
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -106,3 +106,17 @@ exports.app_details = [
       .catch(next);
   },
 ];
+
+
+exports.app_delete = [
+  ensureLoggedIn('/login'),
+  (req, res, next) => {
+    const app_id = req.params.app_id;
+
+    api.apps.delete(app_id)
+      .then(() => {
+        res.redirect(routes.url_for('apps.list'));
+      })
+      .catch(next);
+  }
+];
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -111,7 +111,7 @@ exports.app_details = [
 exports.app_delete = [
   ensureLoggedIn('/login'),
   (req, res, next) => {
-    const app_id = req.params.app_id;
+    const app_id = req.params.id;
 
     api.apps.delete(app_id)
       .then(() => {
--- a/app/apps/routes.js
+++ b/app/apps/routes.js
@@ -7,5 +7,5 @@ module.exports = [
   { name: 'create', method: 'POST', pattern: '/apps/create', handler: handlers.create_app },
   { name: 'list_user_apps', pattern: '/user/:id/apps', handler: handlers.list_user_apps },
   { name: 'details', pattern: '/apps/:id', handler: handlers.app_details },
-  { name: 'delete', method: 'POST', pattern: '/apps/:app_id/delete', handler: handlers.app_delete },
+  { name: 'delete', method: 'POST', pattern: '/apps/:id/delete', handler: handlers.app_delete },
 ];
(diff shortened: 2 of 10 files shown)


app/apps/routes.js
```diff
@@ -1,10 +1,11 @@
...
-  {name: 'list', pattern: '/apps', handler: handlers.list_apps},
-  {name: 'create', method: 'POST', pattern: '/apps/create', handler: handlers.create_app},
-  {name: 'list_user_apps', pattern: '/user/:id/apps', handler: handlers.list_user_apps},
-  {name: 'details', pattern: '/apps/:id', handler: handlers.app_details},
+  { name: 'new', pattern: '/apps/new', handler: handlers.new_app },
+  { name: 'list', pattern: '/apps', handler: handlers.list_apps },
+  { name: 'create', method: 'POST', pattern: '/apps/create', handler: handlers.create_app },
+  { name: 'list_user_apps', pattern: '/user/:id/apps', handler: handlers.list_user_apps },
+  { name: 'details', pattern: '/apps/:id', handler: handlers.app_details },
+  { name: 'delete', method: 'POST', pattern: '/apps/:app_id/delete', handler: handlers.app_delete },
```
xoen: I would call the param in the URL just `:id` for consistency with the other endpoints.

xoen:
Look good to me 👍 

Only two things to node are:
 - I'd rename the param to just `:id` from `:app_id` for consistency
 - we could probably write some tests

app/apps/handlers.js
```diff
@@ -106,3 +106,17 @@ exports.app_details = [
...
   },
 ];
+
+
+exports.app_delete = [
+  ensureLoggedIn('/login'),
+  (req, res, next) => {
+    const app_id = req.params.app_id;
+
+    api.apps.delete(app_id)
```
andyhd: Don't really need to set a variable here, just do `api.apps.delete(req.params.app_id)`

app/assets/javascripts/modules/confirm.js
```diff
@@ -2,6 +2,7 @@
 
 moj.Modules.jsConfirm = {
   confirmClass: 'js-confirm',
+  confirmMessage: 'Are you sure?',
```
andyhd: Not important, but should this message be a bit more specific, like "Are you sure you want to delete this thing?"

axemonkey: Could be. Or, actually, it could be moved out of the JS and into a `data-` attribute so that the module can be used on any confirmable button/action.

app/templates/apps/details.html
```diff
@@ -95,9 +95,9 @@ <h2 class="heading-medium">App data sources</h2>
 <hr>
 
 <!-- only show if signed in user has admin rights to this app -->
-<p class="clearfix align-right">
-  <a class="button button-secondary js-confirm" href="/apps/delete/{{ app.id }}/{{ user_id }}">Delete app</a>
-</p>
+<form action="{{ url_for('apps.delete', {app_id: app.id}) }}" method="post" class="clearfix">
```
xoen: I think you may also need the `inline-form` class here as well.

axemonkey: Not actually needed since there is only one button here, so it's fine to be a right-floated block.

xoen: I see 🙇 I didn't realise that, it was needed in my case as I had two buttons side-by-side (I added these classes in my PR so once this is merged buttons will look awesome 🌟 )

PR #59, opened 2017-11-20
tidying up templates and adding missing buttons

axemonkey:
## What

* hide debug dumps inside `<details>` tags
* add "Create new x" buttons for apps and buckets on list pages
* tidy up some abstruse conditionals
* red delete buttons!


app/templates/users/details.html
```diff
@@ -1,91 +1,107 @@
...
+</div>
+
+
+<h2>User's app data sources</h2>
+{% if user.users3buckets.length %}
+  <table class="form-group">
+    <thead>
+      <tr>
+        <th>Name</th>
+        <th>Admin</th>
```
xoen: Not massively important at this stage but we may want to keep wording of `Admin` and `Read/write access` consistent with other templates at some point in the future (specifically with `buckets.details.html` template).

app/templates/users/details.html
```diff
@@ -1,91 +1,107 @@
...
+        <th>Read/write access</th>
+        <td><span class="visuallyhidden">Actions</span></td>
+      </tr>
+    </thead>
+    {% for bucket in user.users3buckets %}
```
xoen: You may want to rename this variable to `users3bucket` to avoid confusion as this is a loop on the access objects, not the actual buckets.

PR #60, opened 2017-11-21
Add `yes_no` template macro/helper function

xoen:
### What

This helps simplify repetitive if/else blocks in the template.

The function by default returns 'Yes' if the passed value is
true or 'No' if it's false.

### Examples

This is useful in the simple cases like when listing which users are
admin of an S3 bucket, e.g.

    <td>{{ yes_no(users3bucket.is_admin) }}</td>

The function behaviour can be customised to check if the value is equal
to something else, e.g. a string.

This is useful when displaying 'Yes'/'No' depending on the access level
for example:

    <th>Has R/W access?</th>
    <td>{{ yes_no(users3bucket.access_level, 'readwrite') }}</td>

(In the example above the function will compare access level with 'readwrite'
which is considered the "true" value)

The strings returned can be customised so for example you can use the helper
to display something else, e.g.

    <td>{{ yes_no(users3bucket.is_admin, true, '(Admin)', '') }}</td>

Because these strings can be customised, the helper can be used to "toggle"
the access level value, e.g.

    <inpyt type="hidden" name="access_level" value="{{ yes_no(users3bucket.access_level, 'readwrite', 'readonly', 'readwrite') }}" />
    <input type="submit" value="{{ yes_no(users3bucket.access_level, 'readwrite', 'Revoke', 'Grant') }} R/W access" />

### DIFF
--- a/app/middleware/template-locals.js
+++ b/app/middleware/template-locals.js
@@ -1,3 +1,19 @@
+/*
+Examples:
+
+ - yes_no(is_admin) // => 'Yes' / 'No'
+ - yes_no(access_level, 'readwrite') // => 'Yes' / 'No'
+ - yes_no(is_admin, true, '(Admin)', '') // => '(Admin)' / ''
+ - yes_no(access, 'readwrite', 'Revoke', 'Grant') // => 'Revoke' / 'Grant'
+*/
+function yes_no(value, true_value=true, yes_str='Yes', no_str='No') {
+  if (value === true_value) {
+    return yes_str;
+  } else {
+    return no_str;
+  }
+}
+
 module.exports = (app, conf, log) => {
   log.info('adding template-locals');
   return (req, res, next) => {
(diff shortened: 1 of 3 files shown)


app/middleware/template-locals.js
```diff
@@ -1,10 +1,27 @@
...
+ - yes_no(is_admin, true, '(Admin)', '') // => '(Admin)' / ''
+ - yes_no(access, 'readwrite', 'Revoke', 'Grant') // => 'Revoke' / 'Grant'
+*/
+function yes_no(value, true_value=true, yes_str='Yes', no_str='No') {
+  if (value === true_value) {
+    return yes_str;
+  } else {
+    return no_str;
+  }
+}
```
andyhd: This should be a [template macro](https://mozilla.github.io/nunjucks/templating.html#macro), not a local function.

PR #62, opened 2017-11-22
Refactor api client

andyhd:
## What

* Refactored control panel API client library to use ES6 classes, with a (sort-of) pluggable authentication mechanism and class methods for HTTP methods. This simplifies the api client use, eg:
  ```javascript
  api.auth.set_token(jwt);
  api.patch(`users/${user_id}`, {username: 'New Name'})
    .then((updated_user) => {
      console.log(updated_user.username);
    });
  ```

* Added `Model` base class and `ModelSet` collection class to convert JSON representations of API models to objects with useful CRUD methods, eg:
  ```javascript
  // create
  User({json: data}).create()
    .then((new_user) => { console.log(new_user.id); });

  // retrieve
  User.get(id)
    .then((user) => ...);

  // update
  User({json: data}).update()
    .then((updated_user) => ...);

  // delete
  User.delete(user_id)
    .then(() => ...);

  // list
  User.list()
    .then((users) => {
      console.log(users[0].username);
      });
    });
  ```

* Added some methods to `User` and `App` models to get associated objects, eg:
  ```javascript
  User.get(andy_id)
    .then((andy) => {
      // this is probably not a useful example, but you can access .apps, .buckets
      // and .users3buckets on a user object
      return andy.apps[0].connect_bucket(bucket_id);
    })
    .then((new_apps3bucket) => {
      console.log(new_apps3bucket.id);
    });
  ```

* Added a handy `exclude` method on `ModelSet` to filter by primary key, eg:
  ```javascript
  const andy_id = 'github|123';
  User.list()
    .then((users) => {
      const everyone_but_andy = users.exclude(andy_id);
    });
  ```

* Added the error stacktrace to the 500 error page if running in dev

## How to review

1. Nothing should be visibly different, except the code and the stacktrace on the error page


app/apps/handlers.js
```diff
@@ -1,122 +1,100 @@
...
 
 
-exports.create_app = [
+exports.create = [
   ensureLoggedIn('/login'),
-  function (req, res, next) {
-    const app = {
+  (req, res, next) => {
+
+    const { url_for } = require('../routes');
```
xoen: If we're using `url_for()` in different places we may want to move this require at the top of the file.

app/apps/handlers.js
```diff
@@ -1,122 +1,100 @@
...
+
+    new App({
       name: req.body.name,
       description: req.body.description,
       repo_url: req.body.repo_url,
       userapps: [],
-    };
+    })
```
xoen: I'm not so sure about these spaces between these statements to be honest as it may look like they're not related to each other.

app/apps/handlers.js
andyhd: `routes` cannot be imported at the top of a `handlers` module, because it depends on all the handlers being loaded to work.

app/apps/handlers.js
andyhd: Yep, I will remove them. It's in the [code standard](https://github.com/airbnb/javascript#whitespace--padded-blocks) too 🙈

app/apps/handlers.js
andyhd: The reason we had trouble doing `const { url_for } = require('../routes')` in a handler is because there is a cyclic dependency: `routes` requires all the handlers, which require `routes`.

app/apps/routes.js
```diff
@@ -2,10 +2,11 @@ const handlers = require('./handlers');
...
-  { name: 'create', method: 'POST', pattern: '/apps/create', handler: handlers.create_app },
-  { name: 'list_user_apps', pattern: '/user/:id/apps', handler: handlers.list_user_apps },
-  { name: 'details', pattern: '/apps/:id', handler: handlers.app_details },
-  { name: 'delete', method: 'POST', pattern: '/apps/:id/delete', handler: handlers.app_delete },
+  { name: 'new', pattern: '/apps/new', handler: handlers.new },
+  { name: 'list', pattern: '/apps', handler: handlers.list },
+  { name: 'create', method: 'POST', pattern: '/apps/create', handler: handlers.create },
+  { name: 'details', pattern: '/apps/:id', handler: handlers.details },
+  { name: 'delete', method: 'POST', pattern: '/apps/:id/delete', handler: handlers.delete },
+  //{ name: 'connect_bucket', method: 'POST', pattern: '/apps/:app_id/buckets', handler: handlers.connect_bucket },
```
xoen: Don't think we have this endpoint anymore, and also not sure what the `update_bucket` (commented) endpoint would do.

app/apps3buckets/handlers.js
```diff
@@ -1,54 +1,55 @@
...
+
+      .then((app) => {
+        return app.connect_bucket(bucket_id);
+      })
 
-    api.apps3buckets.add(apps3bucket)
-      .then(() => {
-        res.redirect(routes.url_for('apps.details', { id: app_id }));
+      .then((_) => {
+        const { url_for } = require('../routes');
```
xoen: See other comment on potentially moving this require on the top of the file if we use `url_for()` in different places.

app/apps/handlers.js
xoen: Oohhh, so this was the problem with this require, it was really weird.

app/apps3buckets/handlers.js
xoen: Ignore me.

app/apps3buckets/handlers.js
```diff
@@ -1,54 +1,55 @@
...
 
-    const apps3bucket = {
-      app: app_id,
-      s3bucket: bucket_id,
-      access_level: 'readonly',
-    };
+    App.get(app_id)
+
+      .then((app) => {
+        return app.connect_bucket(bucket_id);
```
xoen: I would probably prefer something like `app.grant_bucket_access()`.

app/apps/routes.js
andyhd: They are not supposed to be there, I will remove them

app/apps3buckets/handlers.js
andyhd: We can't do that, as explained

app/buckets/handlers.js
```diff
@@ -23,71 +18,51 @@ exports.list_buckets = [
...
+    Promise.all([Bucket.get(req.params.id), App.list(), User.list()])
+
+      .then(([bucket, apps, users]) => {
+        res.render('buckets/details.html', {
           bucket: bucket,
-          apps_options: get_apps_options(bucket, all_apps),
-          users_options: get_users_options(bucket, all_users),
-        };
-        res.render('buckets/details.html', template_args);
+          apps_options: apps.exclude(bucket.apps),
```
xoen: This code looks tidier 👍

app/api-client.js
```diff
@@ -1,170 +1,291 @@
-const config = require('../app/config');
+const config = require('./config');
```
xoen: Just thinking aloud, we may want to have a small `api-client` file.
And have all these classes in their own file under `apis` or even `models`, just to keep things small and sweet.

(we don't need to do it or do it now, just thinking about it)

app/api-client.js
```diff
@@ -1,170 +1,291 @@
...
+
+
+class APIClient {
+  constructor(conf) {
+    this.base_url = conf.base_url;
+    this.auth = null;
+  }
+
+  request(endpoint, method = 'GET', body = null) {
+    const headers = {};
```
xoen: Watch out here as you're declaring this as a constant but adding to it later.

I think this may still be possible because of the way JS objects work but it still seems weird.

app/api-client.js
```diff
@@ -1,170 +1,291 @@
...
-      return request(override_defaults(options));
+      const options = {
+        method,
+        uri: this.endpoint_url(endpoint),
+        headers,
+        body,
+        json: true,
+      };
+      return request(options)
+        .then((result) => {
```
xoen: So we're resolving the request promise here. Does it mean this `request()` method will be blocking?

I still need to check the rest of the file...sorry if it's a silly question.

app/api-client.js
```diff
@@ -1,170 +1,291 @@
...
-      config.api.username + ':' + config.api.password).toString('base64');
+const model_proxy = {
+  get: (model, property) => {
+    if (Reflect.has(model, property)) {
+      return Reflect.get(model, property);
+    }
+    if (model.data && property in model.data) {
+      return model.data[property];
+    }
+    return undefined;
```
xoen: I wonder if raise an exception here would be better than return `undefined`.
I'm still wrapping my head around this proxy thing but I imagine if trying to read a property which is not there it may be some kind of bug.

app/api-client.js
```diff
@@ -1,170 +1,291 @@
...
+
+exports.api = api;
+
 
-  function basic_auth() {
-    return 'Basic ' + new Buffer(
-      config.api.username + ':' + config.api.password).toString('base64');
+const model_proxy = {
+  get: (model, property) => {
+    if (Reflect.has(model, property)) {
```
xoen: Not familiar with JS `Reflect`, but I get the gist of it.
Is this equivalent to

```js
if (property in model) {
  return model[property];
}
```

Similar to what you do in following `if` statement. Maybe there is some subtle difference I'm missing.

app/api-client.js
```diff
@@ -1,170 +1,291 @@
...
+  constructor(Model, data = []) {
+    super(...data.map(obj => new Model(obj)));
+    this.model = Model;
   }
 
-  function jwt_auth() {
-    return `JWT ${token}`;
+  exclude(other) {
+    let others = other;
+    if (other instanceof this.model) {
```
xoen: You may be able to get rid of the `if` by using the `Array` constructor with the spread, with something like `const others = Array(...other)` (untested but should do).

app/api-client.js
andyhd: `const` means you can't reassign to the same variable name, not that the value is immutable.

app/api-client.js
andyhd: `.then()` actually returns another Promise, that's why you can chain `.then()` and `.catch()`.

app/api-client.js
andyhd: What this code does is similar to Ruby's method_missing.

app/api-client.js
andyhd: I agree, in Python it would raise an AttributeError. But Javascript returns `undefined`, eg:
```javascript
const foo = {};
console.log(foo.bar); // "undefined"
```

app/api-client.js
```diff
@@ -1,170 +1,291 @@
...
-  };
+  create() {
+    return api.post(this.constructor.endpoint, this.data)
+      .then(data => new this.constructor(data));
+  }
 
-  api.list_apps = function () {
-    return api_request({ endpoint: 'apps' });
-  };
+  replace() {
```
xoen: Is this `replace()` method used anywhere? Also, what is it doing? Is it just to make a `PUT` request to the API? We may not need it (yet?).

app/api-client.js
andyhd: Nope, that would fail if `other` is not an iterable.

app/api-client.js
andyhd: I agree, it needs to be split up. `APIClient` and `JWTAuth` don't belong with `Model` and `ModelSet` and the model classes should also be separate.

app/api-client.js
andyhd: Yeah, it's not used anywhere yet.

app/api-client.js
xoen: Sorry, I tried this here and it worked but maybe I'm missing the context here: https://jsfiddle.net/#&togetherjs=voaHu070qF

app/api-client.js
```diff
@@ -1,170 +1,291 @@
...
-
-  api.users3buckets = {
-    add: api.add_users3bucket,
-    update: api.update_users3bucket,
-    delete: api.delete_users3bucket,
-  };
-
-  return api;
-})();
+api.connect_bucket_to_app = apps3bucket => api.post('apps3buckets', body = apps3bucket);
```
xoen: There is something unsettling about this line of code (it may be just me being tired!).

I think as the `api.post()` has side effect and it's not just mapping parenthesis and curly braces may help (I think Airbnb coding standard says something about it).

Also, I was about to say something about this function name but it seems like it's not used anywhere so we may just delete this line.

andyhd: Yep, that's right - it should have curly braces - and it's not used anywhere. I will delete it.

PR #66, opened 2017-11-27
Show users on app page, grant and revoke user admin access to app

axemonkey:
## What

* Show an app's admin users on the app details page
* Allow app admin users to grant/revoke admin access to other users
* Tone down the number of red buttons a bit

Trello [here](https://trello.com/c/YNcvOfMA/465-show-users-on-app-details-page-1) and [here](https://trello.com/c/WMnNqhLq/525-app-admins-can-grant-revoke-app-admin-access-to-other-users-2).

### DIFF
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -49,8 +49,7 @@ exports.list = (req, res, next) => {
 exports.details = (req, res, next) => {
   Promise.all([App.get(req.params.id), Bucket.list(), User.list()])
     .then(([app, buckets, users]) => {
-      const user_ids = app.users.map(user => user.auth0_id);
-      const current_user_is_app_admin = user_ids.includes(req.user.user_id);
+      const current_user_is_app_admin = app.has_admin(req.user.user_id);
       res.render('apps/details.html', {
         app,
         buckets_options: buckets.exclude(app.buckets),
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -1,6 +1,5 @@
 const { App, Bucket, User } = require('../models');
 
-
 exports.new = (req, res, next) => {
   Bucket.list()
     .then((buckets) => {
@@ -50,10 +49,14 @@ exports.list = (req, res, next) => {
 exports.details = (req, res, next) => {
   Promise.all([App.get(req.params.id), Bucket.list(), User.list()])
     .then(([app, buckets, users]) => {
+      const user_ids = app.data.userapps.map(userapp => userapp.user.auth0_id);
+      const current_user_is_app_admin = user_ids.includes(req.user.user_id);
       res.render('apps/details.html', {
         app,
         buckets_options: buckets.exclude(app.buckets),
         users,
+        users_options: users.exclude(app.users),
+        current_user_is_app_admin,
       });
     })
     .catch(next);
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -49,7 +49,7 @@ exports.list = (req, res, next) => {
 exports.details = (req, res, next) => {
   Promise.all([App.get(req.params.id), Bucket.list(), User.list()])
     .then(([app, buckets, users]) => {
-      const user_ids = app.data.userapps.map(userapp => userapp.user.auth0_id);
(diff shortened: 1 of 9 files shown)


app/apps/handlers.js
```diff
@@ -50,10 +49,14 @@ exports.list = (req, res, next) => {
 exports.details = (req, res, next) => {
   Promise.all([App.get(req.params.id), Bucket.list(), User.list()])
     .then(([app, buckets, users]) => {
+      const user_ids = app.data.userapps.map(userapp => userapp.user.auth0_id);
```
xoen: As discussed in Slack, we currently don't use the UserApp relationship for permission-checking but we may start at some point.

That relationship also has an `is_admin` field which should potentially give you not only access to an app but also ability to perform potentially destructive operations.

That field is ignored in this case despite the variable name called `current_user_is_app_admin`, this may be alright for now but it may require tweaking in the future.

xoen: e.g. a user whose `UserApp` record has the `is_admin` flag set to `false` can see the buttons to add/remove "user<->app" relationships.

axemonkey: True. At the moment there is no difference between a user connected to an app who is "an admin" and one who isn't. When connecting a user to an app the `is_admin` field is set to true, so it's always going to be true. That may change.

axemonkey: Yes. At the moment the `is_admin` flag is completely superfluous.

app/models.js
```diff
@@ -31,6 +35,25 @@ class App extends Model {
...
+
+  grant_user_access(user, access_level = 'readonly', is_admin = false) {
+    if (!['readonly', 'readwrite'].includes(access_level)) {
+      throw new Error(`Invalid access_level "${access_level}"`);
+    }
+
+    let user_id = user;
+
+    if (typeof user === 'function' && user.prototype.constructor === User) {
+      user_id = user.id;
```
xoen: Does this work? I don't think the `User` or its base class `Model` have an `id` property (I may be wrong!). You may need to change this to `user.pk` or `user.auth0_id`.

axemonkey: Hmmm. Honestly, I copied this from `grant_bucket_access` and didn't overthink it. The only place this gets called sends in a plain `user_id` so this will never get called. I'll strip it out.

app/apps/handlers.js
andyhd: Don't access `app.data.userapps`, use `app.users`, eg:
```javascript
const user_ids = app.users.map(user => user.auth0_id);
```

andyhd: Oh, there isn't one. But you can add it pretty easily in `app/models.js`. Eg:
```javascript
  get users() {
    return new ModelSet(User, this.data.userapps.map(ua => ua.user));
  }
```

axemonkey:
Nah, that gets passed in from `<option value="{{ user.auth0_id }}">{{ user.name }}</option>` into `req.body`

app/apps/handlers.js
```diff
@@ -50,10 +49,14 @@ exports.list = (req, res, next) => {
...
+      const current_user_is_app_admin = user_ids.includes(req.user.user_id);
```
andyhd: This code might be better placed in the User model, eg:
```javascript
const current_user_is_app_admin = app.has_admin(req.user.user_id)
```

xoen: 👍

app/models.js
```diff
@@ -36,14 +36,14 @@ class App extends Model {
     }).create();
   }
 
-  grant_user_access(user, access_level = 'readonly', is_admin = false) {
+  grant_user_access(user_id, access_level = 'readonly', is_admin = false) {
```
xoen: Nice 👍

app/apps/handlers.js
axemonkey: ❓

axemonkey:
The has admin stuff is hardcoded to return true for now, since otherwise superusers would be prevented from performing these operations in the Control Panel. A TODO is added to return the real value once user perms have been implemented.

PR #68, opened 2017-11-29
Create new bucket and app at the same time

axemonkey:
## What

Enable user to create a new empty bucket at the same time as creating a new app.

## How to review

1. log in
2. click "Create new app"
3. enter name and repo URL
4. select "Create a new datasource" in the radio buttons
5. enter a name (that doesn't already exist) for the datasource
6. click create
7. you should see the app details page with your new datasource created and connected to the new app

## TODO

Check for datasource name collisions – probably applies to app names as well

### DIFF
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -11,7 +11,6 @@ exports.new = (req, res, next) => {
     .catch(next);
 };
 
-
 exports.create = (req, res, next) => {
   const app = new App({
     name: req.body.name,
@@ -31,14 +30,17 @@ exports.create = (req, res, next) => {
 
   Promise.all([app.create(), create_bucket])
     .then(([created_app, created_bucket]) => {
-      if (created_bucket) {
-        created_app.grant_bucket_access(created_bucket.id, 'readonly');
-      } else if(req.body['new-app-datasource'] === 'select') {
-        created_app.grant_bucket_access(req.body['select-existing-datasource'], 'readonly');
-      }
+      let grant_access = Promise.resolve(null);
 
-      const { url_for } = require('../routes'); // eslint-disable-line global-require
-      res.redirect(url_for('apps.details', { id: created_app.id }));
+      if (created_bucket) {
+        grant_access = created_app.grant_bucket_access(created_bucket.id, 'readonly');
+      } else if (req.body['new-app-datasource'] === 'select') {
+        grant_access = created_app.grant_bucket_access(req.body['select-existing-datasource'], 'readonly');
+      }
+      grant_access.then(() => {
+        const { url_for } = require('../routes'); // eslint-disable-line global-require
+        res.redirect(url_for('apps.details', { id: created_app.id }));
+      });
     })
     .catch((err) => {
       if (err.statusCode === 400) {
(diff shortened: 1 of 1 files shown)


app/apps/handlers.js
```diff
@@ -20,12 +20,25 @@ exports.create = (req, res, next) => {
...
+  let new_bucket;
+
+  if(req.body['new-app-datasource'] === 'create') {
+    new Bucket({
+      name: req.body['new-datasource-name']
+    }).create()
+      .then((created_bucket) => {
+        new_bucket = created_bucket;
+      });
+  }
```
andyhd: This might not work - the `Bucket.create` promise here might not resolve before the `app.create` promise below.

You could nest the promises, eg:
```javascript
app.create()
  .then((created_app) => {
    ...
    if (req.body['new-app-datasource'] === 'create') {
      new Bucket({ name: req.body['new-datasource-name'] }).create()
        .then((created_bucket) => {
          created_app.grant_bucket_access(created_bucket.id, 'readonly');
        });
    }
  });
```

Or you could run them in parallel, eg:
```javascript
const create_bucket = new Promise((resolve, reject) => {
  if (req.body['new-app-datasource'] === 'create') {
    resolve(new Bucket({...}).create());
  } else {
    resolve(null);
  }
});
Promise.all([app.create(), create_bucket])
  .then([created_app, created_bucket]) => {
    if (created_bucket) {
      created_app.grant_bucket_access(created_bucket.id, 'readonly');
    }
    ...
  });
```

app/apps/handlers.js
```diff
@@ -20,12 +20,25 @@ exports.create = (req, res, next) => {
...
 
   app.create()
     .then((created_app) => {
       if(req.body['new-app-datasource'] === 'select') {
         created_app.grant_bucket_access(req.body['select-existing-datasource'], 'readonly');
       }
+      if(req.body['new-app-datasource'] === 'create') {
+        created_app.grant_bucket_access(new_bucket.data.id, 'readonly');
```
andyhd: You can access `new_bucket.id` instead of `new_bucket.data.id` - `data` is meant to be sort-of private.

app/apps/handlers.js
axemonkey: 👍  Thanks, will give the latter a go.

andyhd:
`grant_bucket_access` also returns a Promise, so you need to redirect after it resolves

app/apps/handlers.js
```diff
@@ -19,18 +19,26 @@ exports.create = (req, res, next) => {
...
 
-  app.create()
-    .then((created_app) => {
-      if(req.body['new-app-datasource'] === 'select') {
+  Promise.all([app.create(), create_bucket])
+    .then(([created_app, created_bucket]) => {
+      if (created_bucket) {
+        created_app.grant_bucket_access(created_bucket.id, 'readonly');
+      } else if(req.body['new-app-datasource'] === 'select') {
         created_app.grant_bucket_access(req.body['select-existing-datasource'], 'readonly');
```
andyhd: `grant_bucket_access` also returns a Promise, so you'll need to do something like:
```javascript
let grant_access = Promise.resolve(null);
if (created_bucket) {
  grant_access = created_app.grant_bucket_access(...);
} else if (...) {
  grant_access = created_app.grant_bucket_access(...);
}
grant_access.then(() => {
  res.redirect(url_for(...));
}
```

PR #70, opened 2017-11-30
When creating an app, become admin user of that app

axemonkey:
## What

When a user creates an app in the Control Panel they should automatically be added as an admin user of that app

## How to review

1. log in
2. create an app
3. you should be listed as an admin user of that app on the app details page

### DIFF
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -1,4 +1,4 @@
-const { App, Bucket, User } = require('../models');
+const { App, Bucket, User, UserApp } = require('../models');
 
 exports.new = (req, res, next) => {
   Bucket.list()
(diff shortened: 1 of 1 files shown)


app/apps/handlers.js
```diff
@@ -37,10 +37,12 @@ exports.create = (req, res, next) => {
       } else if (req.body['new-app-datasource'] === 'select') {
         grant_access = created_app.grant_bucket_access(req.body['select-existing-datasource'], 'readonly');
       }
-      grant_access.then(() => {
-        const { url_for } = require('../routes'); // eslint-disable-line global-require
-        res.redirect(url_for('apps.details', { id: created_app.id }));
-      });
+      grant_access
+        .then(() => created_app.grant_user_access(req.user.user_id, 'readwrite', true))
```
andyhd: This needs curly braces - grant_user_access has side-effects.

app/apps/handlers.js
```diff
@@ -1,4 +1,4 @@
-const { App, Bucket, User } = require('../models');
+const { App, Bucket, User, UserApp } = require('../models');
```
andyhd: This isn't used anywhere.

axemonkey: Doh

app/apps/handlers.js
axemonkey: Doh

PR #71, opened 2017-12-01
Prefix repo url, get repo description from github

axemonkey:
## What

* Make stored URL from i) predetermined Github domain prefix, ii) org selected from dropdown, iii) slug entered as free text
* Store the complete URL in a hidden field and the slug in the old `name` field
* On entering a complete URL, check Github for repo existence, and if it exists,
* Get the description and store it in a hidden field

## How to review

1. Check out, run, log in
2. Create a new app in the control panel
3. See that when you complete a repo URL the application checks Github and gets a description if there is one
4. See that if you create an app, the correct details are stored and shown in the apps list

### DIFF
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -4,7 +4,12 @@ exports.new = (req, res, next) => {
   Bucket.list()
     .then((buckets) => {
       res.render('apps/new.html', {
-        prefix: `${process.env.ENV}-`,
+        repo_prefix: 'https://github.com',
+        orgs: [
+          'moj-analytical-services',
+          'ministryofjustice',
+        ],
+        bucket_prefix: `${process.env.ENV}-`,
         buckets,
       });
     })
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -1,14 +1,12 @@
 const { App, Bucket, User } = require('../models');
+const { repos } = require('../config');
 
 exports.new = (req, res, next) => {
   Bucket.list()
     .then((buckets) => {
       res.render('apps/new.html', {
-        repo_prefix: 'https://github.com',
-        orgs: [
-          'moj-analytical-services',
-          'ministryofjustice',
-        ],
+        repo_prefix: repos.host,
+        orgs: repos.orgs,
         bucket_prefix: `${process.env.ENV}-`,
         buckets,
       });
--- a/app/assets/javascripts/modules/repo-description.js
+++ b/app/assets/javascripts/modules/repo-description.js
@@ -1,0 +1,99 @@
+'use strict';
+
+moj.Modules.repoDescription = {
+  repoPrefix: 'https://github.com',
+  apiUrlPrefix: 'https://api.github.com/repos',
+  orgSelectName: 'repo_org',
+  repoSlugInputName: 'name',
+  repoUrlHiddenInputName: 'repo_url',
+  descriptionInputName: 'description',
+
+  messages: {
+    checking: 'Looking up repo...',
+    not_found: 'Repo not found',
+    found: 'Repo found'
+  },
+
+  init: function() {
+    var self = this;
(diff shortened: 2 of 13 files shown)


app/apps/handlers.js
```diff
@@ -4,7 +4,12 @@ exports.new = (req, res, next) => {
   Bucket.list()
     .then((buckets) => {
       res.render('apps/new.html', {
-        prefix: `${process.env.ENV}-`,
+        repo_prefix: 'https://github.com',
+        orgs: [
+          'moj-analytical-services',
+          'ministryofjustice',
+        ],
```
andyhd: This list should be in config

andyhd: Probably the `repo_prefix` too

app/assets/javascripts/modules/repo-description.js
```diff
@@ -0,0 +1,99 @@
...
+
+    $('#repo-check, #repo-checking, .spinner').removeClass('js-hidden');
+    $('#repo-checking p').removeClass('success error').text(self.messages.checking);
+    $('#repo-results').addClass('js-hidden');
+
+    $.ajax({
+      method: 'GET',
+      dataType: 'jsonp',
+      callback: 'callback',
+      url: apiUrl,
```
andyhd: Pretty sure this is only going to work for public repos, right?

andyhd: I think it would be better to do this server-side and use the current user's github access token to request the data from the API

andyhd: But maybe that's a new ticket, rather than a fix for this PR

axemonkey: Fine by me ¯\_(ツ)_/¯

app/apps/handlers.js
axemonkey: 👍

app/assets/javascripts/modules/repo-description.js
```diff
@@ -0,0 +1,99 @@
...
+
+    $('#repo-check, #repo-results').removeClass('js-hidden');
+    $('#repo-checking .spinner').addClass('js-hidden');
+
+    if(data.id) {
+      $('#' + self.descriptionInputName).val(description);
+      if(description) {
+        $('#repo-' + self.descriptionInputName).text(description);
+      } else {
+        $('#repo-' + self.descriptionInputName).text('None provided');
```
xoen: I just started to have a look at this so apologies if I'm missing something.

Here the `None provided` string will be the app description if the repository description is blank right? I would personally prefer that to be simply an empty string in that case (also that would make the code slightly simpler), instead of using an arbitrary string.

(this is not a big deal obviously)

axemonkey: No, it doesn't change the value of `description`, so an empty string would be stored.

app/templates/apps/new.html
```diff
@@ -4,39 +4,54 @@
...
+<div class="text">
+  <p class="lede form-group">After <a href="https://moj-analytical-services.github.io/platform_user_guidance/deploying-a-shiny-app.html">creating an app in Github</a>, use this form to register your app and connect it to sensitive data stored in S3.</p>
+</div>
+
 
+<form action="{{ url_for('apps.create') }}" method="post">
   <div class="form-group {% if errors.name %} form-group-error {% endif %}">
-    <label for="name" class="form-label">Name</label>
+    <label for="name" class="heading-medium form-label">
+      Github repo url
```
xoen: Minor wording question, shouldn't be 'URL' instead of 'url'?

axemonkey: 👍

xoen: /me waiting for punch from the side :P

app/assets/javascripts/modules/repo-description.js
xoen: 👍

PR #73, opened 2017-12-05
Build Docker image faster

xoen:
- Don't include generated things into docker context (`node_modules` and
  `static` are generated inside the container, no reason to pass them in)
- Moved addition of `/app` and `/bin` AFTER `npm_install`, this is slow
  and only needs to be run again if `packages.json` changes, not if any of
  the application's files change.

### See
- https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#use-a-dockerignore-file
- article with a similar example of a Dockerfile for a node app https://ngeor.wordpress.com/2017/11/15/cd-with-helm-part-1-dockerize-it/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,15 +4,13 @@ MAINTAINER Andy Driver <andy.driver@digital.justice.gov.uk>
 
 WORKDIR /home/cpanel
 
+ADD package.json package.json
+RUN npm install
+
 ADD app app/
 ADD bin bin/
-ADD package.json package.json
-
-RUN npm install
-ADD node_modules node_modules/
 
 RUN npm run-script collect-static
-ADD static static/
 
 ENV EXPRESS_HOST "0.0.0.0"
 ENV NODE_RESTART "0"
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,8 @@ WORKDIR /home/cpanel
 ADD package.json package.json
 RUN npm install
 
-ADD app app/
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -4,15 +4,13 @@ MAINTAINER Andy Driver <andy.driver@digital.justice.gov.uk>
...
 
-ADD app app/
-ADD bin bin/
 ADD package.json package.json
-
 RUN npm install
-ADD node_modules node_modules/
+
+ADD app app/
+ADD bin bin/
```
andyhd: I like this - one point: `bin/` will change far less frequently than `app/`, so put it first and we'll benefit from the cache even more

Dockerfile
```diff
@@ -4,15 +4,13 @@ MAINTAINER Andy Driver <andy.driver@digital.justice.gov.uk>
...
 
 RUN npm run-script collect-static
-ADD static static/
```
andyhd: I think we might need `static/` because possibly not all static files are generated.

Dockerfile
xoen: Good point, will change 👍

Dockerfile
xoen: Aren't they? I tried to delete `static` and run `npm run-script collect-static` and as far as I could tell `git status` didn't see any change (that's why I assumed everything in `static` was generated off other files)

A question for our own JS/node ninja @axemonkey probably :)

PR #75, opened 2017-12-05
Kubernetes API client and models

andyhd:
## What

* Add Kubernetes API client to abstract making requests to the correct endpoints
* Add `Deployment` and `Pod` models to make dealing with Kubernetes API similar to existing API
* Set the namespace for Kubernetes API calls using the logged in user's username
* Allows listing deployments (tools), eg:
  ```javascript
  Deployment.list()
    .then((tools) => {
      tools.forEach((tool) => {
        console.log(tool.metadata.labels.app); // rstudio
      });
    });
  ```
* Allows restarting RStudio, eg:
  ```javascript
  Deployment.get('andyhd-rstudio-rstudio') // Deployment().metadata.name
    .then((rstudio) => {
      return rstudio.restart();
    });
  ```

See [Trello #516](https://trello.com/c/EYuj7Gfs) and [Trello #510](https://trello.com/c/wympY8SG)

### DIFF
--- a/app/api-client.js
+++ b/app/api-client.js
@@ -24,6 +24,8 @@ class JWTAuth {
   }
 }
 
+exports.JWTAuth = JWTAuth;
+
 
 class APIClient {
   constructor(conf) {
@@ -31,7 +33,7 @@ class APIClient {
     this.auth = null;
   }
 
-  request(endpoint, method = 'GET', body = null) {
+  request(endpoint, { method='GET', body=null, params={} } = {}) {
     const headers = {};
 
     if (this.auth) {
@@ -44,11 +46,12 @@ class APIClient {
         uri: this.endpoint_url(endpoint),
         headers,
         body,
+        qs: params,
         json: true,
       };
+      log.debug(`${method} ${options.uri}`);
       return request(options)
         .then((result) => {
-          log.debug(`${method} ${options.uri}`);
           // console.dir(result);
           return result;
         });
@@ -57,24 +60,24 @@ class APIClient {
     }
   }
 
-  get(endpoint) {
-    return this.request(endpoint);
+  get(endpoint, params = {}) {
+    return this.request(endpoint, { params });
   }
 
   post(endpoint, body = '') {
-    return this.request(endpoint, 'POST', body);
+    return this.request(endpoint, { method: 'POST', body });
   }
 
-  delete(endpoint) {
-    return this.request(endpoint, 'DELETE');
+  delete(endpoint, params = {}) {
+    return this.request(endpoint, { method: 'DELETE', params });
   }
 
   patch(endpoint, body = '') {
-    return this.request(endpoint, 'PATCH', body);
+    return this.request(endpoint, { method: 'PATCH', body});
   }
 
   put(endpoint, body = '') {
-    return this.request(endpoint, 'PUT', body);
+    return this.request(endpoint, { method: 'PUT', body});
   }
 
   endpoint_url(endpoint) {
@@ -91,6 +94,7 @@ class APIClient {
   }
 }
 
+exports.APIClient = APIClient;
 
 const api = new APIClient(config.api);
 api.auth = new JWTAuth();
--- a/app/k8s-api-client.js
+++ b/app/k8s-api-client.js
@@ -1,0 +1,27 @@
+const config = require('./config');
+const { APIClient, JWTAuth } = require('./api-client');
+const url = require('url');
+
+
+class KubernetesAPIClient extends APIClient {
+  endpoint_url(endpoint, namespace = 'default') {
+    if (!endpoint) {
+      throw new Error('Missing endpoint');
+    }
+
+    const api = {
+      'deployments': 'apis/apps/v1beta2',
+      'pods': 'api/v1',
+    }[endpoint];
+
+    return url.resolve(this.base_url, `k8s/${api}/namespaces/${namespace}/${endpoint}`);
+  }
+}
+
+
+exports.KubernetesAPIClient = KubernetesAPIClient;
+
+const api = new KubernetesAPIClient(config.api);
+api.auth = new JWTAuth();
+
+exports.api = api;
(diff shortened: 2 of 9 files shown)


app/api-client.js
```diff
@@ -24,14 +24,16 @@ class JWTAuth {
...
+
 
 class APIClient {
   constructor(conf) {
     this.base_url = conf.base_url;
     this.auth = null;
   }
 
-  request(endpoint, method = 'GET', body = null) {
+  request(endpoint, { method='GET', body=null, params={} } = {}) {
```
andyhd: This is odd looking, but it is the only way I found to get named function parameters with default values.

xoen: Maybe I'm missing something here, could this be tackled by having something like?

```javascript
  request(endpoint, args = {}) {
    const method = args.method || 'GET';
    const params = args.params || {};
    const body = args.body || null;
    // ...
```

Obviously not a big deal, it's just a bit odd looking as you said (is this restructuring here?).

app/k8s-api-client.js
```diff
@@ -0,0 +1,36 @@
+const config = require('./config');
+const { APIClient, JWTAuth } = require('./api-client');
+const url = require('url');
+
+
+class KubernetesAPIClient extends APIClient {
```
xoen: nice 👍

app/k8s-api-client.js
```diff
@@ -0,0 +1,36 @@
...
+  endpoint_url(endpoint, namespace = undefined) {
+    if (!endpoint) {
+      throw new Error('Missing endpoint');
+    }
+
+    let ns = namespace;
```
xoen: Couldn't this logic be replaced with something like:

```javascript
let ns = this.namespace || 'default';
```

Another couple of things to bear in mind:
1) We may want to perform certain operation (as a superuser) in other users' namespaces, e.g. Robin wants to restart a user rstudio instance.
2) We may use this client to make requests for non-namespaces things, e.g. get the list of namespaces, although maybe not yet (just a thought)

app/api-client.js
andyhd: That code is equivalent, but way more verbose - the method I'm using is correct, I just thought it might raise a question or two.

app/k8s-model.js
```diff
@@ -0,0 +1,22 @@
+const { Model, ModelSet } = require('./base-model');
+const { api } = require('./k8s-api-client');
+
+
+class K8sModel extends Model {
```
xoen: Is this model actually re-using anything from the original `Model`?

e.g. if for whatever reason we call `K8sModel.create()` it would use the other api, maybe `Model` and `K8sModel` don't have much in common? (I may be wrong/miss something!)

xoen: Oh, all these `this.prototype.constructor` are referring to the parent class here.

app/k8s-api-client.js
andyhd: No, it's more like `let ns = namespace || this.namespace || 'default';` - maybe I should change it to that.

andyhd: The `namespace` parameter can be overridden for superusers accessing other users' namespaces, although probably the easiest way would be to set `api.namespace` before making the request.

andyhd: This code might need to be revisited if we have to list namespaces, but YAGNI

app/k8s-model.js
andyhd: It uses the constructor

andyhd: I didn't bother overriding the methods we don't use yet. We might need to override `.create` if we use it in future.

app/middleware/api.js
```diff
@@ -1,9 +1,21 @@
+function sanitize_username(username) {
```
xoen: I'm happy with having this in a function and obviously good to lowercase the username...although, not so sure about the other transformations, doesn't this mean that **potentially** we make a request to a wrong namespace?

andyhd: [Kubernetes namespaces have to be valid DNS labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#namespaces-and-dns). This is the same way we santize URLs everywhere else (or it should be, if not).

app/models.js
```diff
@@ -133,3 +134,29 @@ class UserApp extends Model {
...
+class Deployment extends K8sModel {
+  static get endpoint() {
+    return 'deployments';
+  }
+
+  get_pods() {
+    return Pod.list({ labelSelector: `app=${this.data.metadata.labels.app}` });
+  }
+
+  restart() {
```
xoen: Nice 👍

app/api-client.js
andyhd: Ok, so it's not the _only_ way.

app/k8s-model.js
andyhd: `this.prototype.constructor` is equivalent to Python `self.__class__`

app/middleware/api.js
andyhd: Technically, we don't have to strip the non-alphanumeric character from the start of the username (L3), because the namespace is always prefixed with `user-`. Also, trimming to 63 chars doesn't take the `user-` prefix into account. I'll have another go at this.

app/k8s-api-client.js
xoen: Agreed.

test/test-k8s-models.js
```diff
@@ -0,0 +1,50 @@
...
+      const ns = 'user-test';
+      const deployments_response = require('./fixtures/apps');
+      const pods_response = require('./fixtures/deployment-pods');
+      const expected = new ModelSet(Pod, pods_response.items);
+
+      const pods_request = mock_api()
+        .get(`/k8s/api/v1/namespaces/${ns}/pods?labelSelector=app%3Drstudio`)
+        .reply(200, pods_response);
+
+      const rstudio = new Deployment(require('./fixtures/deployments').items[0]);
```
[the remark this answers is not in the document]
andyhd: This is the first time this fixture is loaded - you might be confusing `fixtures/deployment-pods` with `fixtures/deployments`?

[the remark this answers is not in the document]
andyhd: Oh, you mean share the fixture across the tests. Yep 👍

PR #76, opened 2017-12-06
homepage for users and superusers

axemonkey:
## What

* Revert to having one template for the homepage for users and superusers
* Show superuser functions (list all users, etc) to superusers only
* Show user's apps and buckets on homepage for all users

## How to review

1. check out, run, log in as superuser
2. see your homepage including "Superuser functions" at the top
3. if you can log in as a non superuser, do so and you will see the same homepage without the superuser functions

superuser homepage
---
screen shot 2017-12-06 at 14 08 48

regular user homepage
---
screen shot 2017-12-06 at 14 09 04


app/apps/handlers.js
```diff
@@ -90,7 +90,11 @@ exports.delete = (req, res, next) => {
   App.delete(req.params.id)
     .then(() => {
       const { url_for } = require('../routes'); // eslint-disable-line global-require
-      res.redirect(url_for('apps.list'));
+      let redirect_to = 'base.home';
+      if(req.user.is_superuser) {
```
andyhd: Missing a space between `if` and `(`

app/base/handlers.js
```diff
@@ -6,11 +6,15 @@ const raven = require('raven');
...
-    res.render('home/superuser.html');
-  } else {
-    res.render('home/user.html');
-  }
+  User.get(req.user.auth0_id)
+    .then((user) => {
+      res.render('base/home.html', {
+        signedInUser: true,
+        user,
+        isSuperuser: req.user.is_superuser,
```
andyhd: The template has a `current_user` variable already available, so instead of passing `isSuperuser`, you could just use `current_user.is_superuser`. Also, you're passing the current user to the template as `user` here, so you could also use `user.is_superuser`

app/templates/base/home.html
```diff
@@ -0,0 +1,29 @@
+{% extends "layouts/one-column.html" %}
+
+{% set page_title = "Hello " + ( current_user.name if current_user ) %}
+{% set heading_prefix = "Your" %}
+
+{% block main_column %}
+
+  {% if isSuperuser %}
```
andyhd: This could be `if user.is_superuser` or `if current_user.is_superuser`

app/templates/users/list.html
```diff
@@ -12,19 +12,23 @@
       <th>Name</th>
       <th>Username</th>
       <th>Email</th>
-      <th><span class="visuallyhidden">Actions</span></th>
+      {% if isSuperuser %}
```
andyhd: 👀

app/templates/users/list.html
```diff
@@ -12,19 +12,23 @@
...
       <tr>
-        <td><a href="{{ url_for('users.details', {id: user.auth0_id}) }}">{{ user.name | default(user.username, true) }}</a></td>
+        <td><a class="{% if current_user.auth0_id == user.auth0_id %}highlight-current{% endif %}"" href="{{ url_for('users.details', {id: user.auth0_id}) }}">{{ user.name | default(user.username, true) }}</a></td>
         <td>{{ user.username }}</td>
         <td>{{ user.email }}</td>
-        <td class="align-right">
-          <a class="button button-secondary" href="{{ url_for('users.edit', {id: user.auth0_id}) }}">Edit</a>
-          <a class="button button-warning js-confirm" href="#">Delete</a>
-        </td>
+        {% if isSuperuser %}
```
andyhd: 👆

axemonkey:
:(

PR #80, opened 2017-12-08
Wire up delete bucket

axemonkey:
## What

Make the various "Delete bucket" buttons work. Decided to change the redirect behaviour to not be based on whether the user us superuser or not, but instead on what page they were on. I.e. if they were on a list of buckets and they delete a bucket, they probably want to stay on that list page. If they are on a bucket details page and they delete, they get returned to their homepage. etc.

## How to review

1. create buckets and delete them
2. 🕺

### DIFF
--- a/app/buckets/handlers.js
+++ b/app/buckets/handlers.js
@@ -49,11 +49,16 @@ exports.bucket_details = (req, res, next) => {
 };
 
 exports.delete = (req, res, next) => {
-  Promise.all([Bucket.get(req.params.id), Bucket.delete(req.params.id)])
-    .then(([bucket]) => {
+  let bucket_name;
+  Bucket.get(req.params.id)
+    .then((bucket) => {
+      bucket_name = bucket.name;
+      return Bucket.delete(bucket.id);
+    })
+    .then(() => {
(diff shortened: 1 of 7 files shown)


app/buckets/handlers.js
```diff
@@ -37,11 +37,24 @@ exports.create_bucket = (req, res) => {
...
         apps_options: apps.exclude(bucket.apps),
         users_options: users.exclude(bucket.users),
+        current_user_is_bucket_admin
       });
     })
     .catch(next);
 };
+
+exports.delete = (req, res, next) => {
+  Promise.all([Bucket.get(req.params.id), Bucket.delete(req.params.id)])
```
andyhd: Doing these in parallel causes a race condition. What if `delete` happens before `get`? Either chain these, eg:
```javascript
Bucket.get(req.params.id)
  .then((bucket) => {
    return Bucket.delete(bucket.id);
  })
  .then(() => {
    ...redirect...
  })
  .catch(next);
```
or, if you change the flash message to not rely on `bucket.name`, you can just do `delete` without the `get`.

axemonkey: True, but I only did the `.get` because I wanted the bucket name. It's nicer. I'll chain them. 👍

PR #81, opened 2017-12-08
Add list and restart tools views

andyhd:
# What

* Added tools routes and handlers for listing tools and restarting
* Added templates for listing tools

## How to review

0. Make sure you have a kubernetes config file at `$HOME/.kube/config`
1. Browse to http://localhost:3000/tools
2. See your RStudio instance listed (if you have one!)
3. Click on the `Restart` button
4. See that your RStudio instance is restarted (should get a 503 when trying to access it for a few seconds, or you can check with `kubectl get pods -n user-{your-username}`).

See [Trello #510](https://trello.com/c/wympY8SG) and [Trello #516](https://trello.com/c/EYuj7Gfs)

### DIFF
--- a/app/k8s-api-client.js
+++ b/app/k8s-api-client.js
@@ -22,16 +22,16 @@ exports.get_namespace = get_namespace;
 
 class KubernetesAPIClient extends APIClient {
   endpoint_url(endpoint, namespace = undefined) {
+    let ns = namespace || this.namespace || 'default';
+
     if (!endpoint) {
       throw new Error('Missing endpoint');
     }
 
-    let ns = namespace || this.namespace || 'default';
-
     const api = {
       'deployments': 'apis/apps/v1beta2',
       'pods': 'api/v1',
-    }[endpoint];
+    }[endpoint.split('/')[0]];
 
     return url.resolve(this.base_url, `k8s/${api}/namespaces/${ns}/${endpoint}`);
   }
--- a/app/models.js
+++ b/app/models.js
@@ -154,8 +154,7 @@ class Deployment extends K8sModel {
   }
 
   get_status() {
-    for (let i in this.status.conditions) {
-      let condition = this.status.conditions[i];
+    for (let condition of this.status.conditions) {
       if (condition.type === 'Available') {
         if (condition.status === 'True') {
           return 'Available';
--- a/app/models.js
+++ b/app/models.js
@@ -141,12 +141,30 @@ class Deployment extends K8sModel {
     return 'deployments';
   }
 
+  static restart(label) {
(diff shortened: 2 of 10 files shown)


app/tools/routes.js
```diff
@@ -0,0 +1,7 @@
+const handlers = require('./handlers');
+
+
+module.exports = [
+  { name: 'list', pattern: '/tools', handler: handlers.list },
+  { name: 'restart', pattern: '/tools/:id/restart', handler: handlers.restart },
```
xoen: I would rename this param to `:name` as this is the tool name to restart, e.g. `rstudio`.

xoen: (I'm working on the endpoint to deploy a tool and the endpoint I have in mind is `POST /tools/:name/deploy`. It's nice to see we thought about similar endpoints)

app/k8s-api-client.js
```diff
@@ -22,16 +22,16 @@ exports.get_namespace = get_namespace;
...
       throw new Error('Missing endpoint');
     }
 
-    let ns = namespace || this.namespace || 'default';
-
     const api = {
       'deployments': 'apis/apps/v1beta2',
       'pods': 'api/v1',
-    }[endpoint];
+    }[endpoint.split('/')[0]];
```
xoen: I probably didn't read the relevant code yet but...why the split here? I've seen both `Deployment` and `Pod` `endpoint()` returns a string without `/`, what am I missing here?

app/models.js
```diff
@@ -141,13 +141,31 @@ class Deployment extends K8sModel {
...
   get_pods() {
     return Pod.list({ labelSelector: `app=${this.data.metadata.labels.app}` });
   }
 
   restart() {
     return Pod.delete_all({ labelSelector: `app=${this.data.metadata.labels.app}` });
   }
+
+  get_status() {
+    for (let i in this.status.conditions) {
```
xoen: Isn't this equivalent to:

```javascript
for (let condition in this.status.conditions) {
  // ...
}
```

app/tools/routes.js
andyhd: Yep

app/k8s-api-client.js
andyhd: This is a bugfix - the endpoint passed in can be something like `deployment/andyhd-rstudio-rstudio`, but that doesn't exist in the `api` lookup object here.

app/models.js
andyhd: No, `for .. in` iterates through the indices (0, 1, 2, ...).
Perhaps you meant `for .. of`, which I didn't realise works with arrays! I will change it.

xoen: Yep, damn you JS :P

PR #83, opened 2017-12-12
Calculate tool status

andyhd:
## What

* Add `display_status` attribute to `Deployment` model which provides a status string based on the `Pod`s associated with the deployment.

## How to review

1. Browse to `http://localhost:3000/tools`
2. See the status of your `rstudio` instance is
   ```
   1. Running
   ```
3. Click `Restart` button
4. See the status of your `rstudio` instance change to something like
   ```
   1. Waiting: ContainerCreating
   2. Running
   ```
5. Refresh your browser and see the status change eventually, possibly to
   ```
   1. Waiting: ContainerCreating
   2. Terminated: ExitCode:0
   ```
   and then eventually back to
   ```
   1. Running
   ```
   (unless there is a problem).

See [Trello #514](https://trello.com/c/r2u3HGhS)

### DIFF
--- a/app/models.js
+++ b/app/models.js
@@ -178,6 +178,37 @@ class Pod extends K8sModel {
   static get endpoint() {
     return 'pods';
   }
+
+  get display_status() {
+    // based on
+    // https://github.com/kubernetes/dashboard/blob/91c54261c6a3d7f601c67a2ccfbbe79f3b6a89f9/src/app/frontend/pod/list/card_component.js#L98
+
+    let containerStatus = this.data.status.containerStatuses;
+
+    if (containerStatus) {
+      let state = containerStatus[0].state;
+
+      if (state.waiting) {
+        return `Waiting: ${state.waiting.reason}`;
+      }
+
+      if (state.terminated) {
+        let reason = state.terminated.reason;
+
+        if (!reason) {
+          if (state.terminated.signal) {
+            reason = `Signal:${state.terminated.signal}`;
+          } else {
+            reason = `ExitCode:${state.terminated.exitCode}`;
+          }
+        }
+
+        return `Terminated: ${reason}`;
+      }
+    }
+
+    return this.data.status.phase;
+  }
 }
 
 exports.Pod = Pod;
--- a/app/base-model.js
+++ b/app/base-model.js
@@ -16,7 +16,7 @@ const model_proxy = {
     if (Reflect.has(model, property)) {
       return Reflect.set(model, property, value);
     }
-    model.data[property] = value;
+    return model.data[property] = value;
   }
 };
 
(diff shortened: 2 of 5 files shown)


app/models.js
```diff
@@ -178,6 +178,45 @@ class Pod extends K8sModel {
...
     return 'pods';
   }
+
+  get display_status() {
+    let containerStatus = this.data.status.containerStatuses;
+    let display_state = 'running';
+    let reason = undefined;
+
+    if (containerStatus) {
+      for (let i = containerStatus.length - 1; i >= 0; i--) {
```
xoen: As far as I can tell you're not using the `i` index so you could want to replace this loop with:

```
for (let status of containerStatus) {
  let state = status.state;
  // ...
```

This has the benefit to be more readable.

... hang on a second, this is looping the statuses in the reverse order, why? (genuine question)

xoen: Are they in some kind of chronological order and you want to only the status in the first element of the array?

Also, as `display_rate` and `reason` are override by the last loop iteration, couldn't you just set these based on the first element of the array?

xoen: I may miss something fundamental here but potentially you can get rid of the look:

```javascript
  get display_status() {
    let containerStatus = this.data.status.containerStatuses;
    let display_state = 'unknown';
    let reason = undefined;

    if (containerStatus) {
      const state = containerStatus[0].state;

      if (state.waiting) {
        display_state = 'waiting';
        reason = state.waiting.reason;
        return `Waiting: ${reason}`;
      }

      if (state.terminated) {
        display_state = 'terminated';
        reason = state.terminated.reason;
        if (!reason) {
          if (state.terminated.signal) {
            reason = `Signal:${state.terminated.signal}`;
          } else {
            reason = `ExitCode:${state.terminated.exitCode}`;
          }
        }
        return `Terminated: ${reason}`;
      }
    }

    return this.data.status.phase;
  }
```

**NOTE**: It seems to work but obviously I may miss something here.

xoen: In fact, you probably can also get rid of `display_state` and other variables :) 

```javascript
  get display_status() {
    const containerStatus = this.data.status.containerStatuses;

    if (containerStatus) {
      const state = containerStatus[0].state;

      if (state.waiting) {
        return `Waiting: ${state.waiting.reason}`;
      }

      if (state.terminated) {
        let reason = state.terminated.reason;
        if (!reason) {
          if (state.terminated.signal) {
            reason = `Signal:${state.terminated.signal}`;
          } else {
            reason = `ExitCode:${state.terminated.exitCode}`;
          }
        }
        return `Terminated: ${reason}`;
      }
    }

    return this.data.status.phase;
  }
```

I would imagine you can squeeze this a bit further as well.

andyhd: This is a straight translation of https://github.com/kubernetes/dashboard/blob/91c54261c6a3d7f601c67a2ccfbbe79f3b6a89f9/src/app/frontend/pod/list/card_component.js#L98 - I kept the reverse iteration because I don't know why they're doing it in Go - I assume a pointless optimisation (it's usually faster to loop down to zero).
I kept the code the same to avoid introducing any bugs.

PR #84, opened 2017-12-13
Add ability to deploy your RStudio instance

xoen:
### What
When a user doesn't have this tool already deployed, a button to `Deploy RStudio` will be shown.

As part of these changes we now also show a link to the tool.

### Why
So that users will be able to deploy RStudio without our intervention.

### Environment variables
Before deploying we need to set the `TOOLS_DOMAIN` environment variable.

### TODO
- [x] Tests
- [x] Possibly update the `docker-compose.yaml`

### IMPORTANT: Depends on
Merge the following PRs before merging this.
- [x] https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/63
- [X] https://github.com/ministryofjustice/analytics-platform-config/pull/22

### Ticket
https://trello.com/c/BbHaXQXt/562-2-cp-app-add-node-endpoint-to-deploy-a-tool

### DIFF
--- a/app/models.js
+++ b/app/models.js
@@ -159,13 +159,12 @@ class Deployment extends K8sModel {
   }
 
   get_status() {
-    for (let condition of this.status.conditions) {
+    for (const condition of this.status.conditions) {
       if (condition.type === 'Available') {
         if (condition.status === 'True') {
           return 'Available';
-        } else {
-          return 'Unavailable';
         }
+        return 'Unavailable';
       }
     }
     return 'Unknown';
@@ -184,17 +183,17 @@ class Pod extends K8sModel {
     // based on
     // https://github.com/kubernetes/dashboard/blob/91c54261c6a3d7f601c67a2ccfbbe79f3b6a89f9/src/app/frontend/pod/list/card_component.js#L98
 
-    let containerStatus = this.data.status.containerStatuses;
+    const containerStatus = this.data.status.containerStatuses;
 
     if (containerStatus) {
-      let state = containerStatus[0].state;
+      const { state } = containerStatus[0];
 
       if (state.waiting) {
         return `Waiting: ${state.waiting.reason}`;
       }
 
       if (state.terminated) {
-        let reason = state.terminated.reason;
+        let { reason } = state.terminated;
 
         if (!reason) {
           if (state.terminated.signal) {
@@ -216,7 +215,6 @@ exports.Pod = Pod;
 
 
 class ToolDeployment {
-
   constructor(data) {
     this.tool_name = data.tool_name;
   }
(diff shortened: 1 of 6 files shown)


xoen:
@axemonkey and @andyhd I don't know if you want to start to have a look at this.

app/models.js
```diff
@@ -212,3 +213,22 @@ class Pod extends K8sModel {
 }
 
 exports.Pod = Pod;
+
+
+class ToolDeployment {
+
```
andyhd: No blank line here - did you run `npm lint`?

app/tools/handlers.js
```diff
@@ -39,3 +45,15 @@ exports.restart = (req, res, next) => {
...
+
+exports.deploy = (req, res, next) => {
+  const { url_for } = require('../routes'); // eslint-disable-line global-require
+
+  new ToolDeployment({tool_name: req.params.name}).create()
+
+  req.session.flash_messages.push(`Deploying '${req.params.name}'...this may take up to 5 minutes`);
+  setTimeout(() => {
+    res.redirect(url_for('tools.list'));
+  }, 2000);
```
andyhd: This timeout is not the best solution - you can't guarantee that the deployment will even start within an arbitrary time. This should explicitly wait for some event, or not wait at all and the tool list page should update dynamically.

Just leaving this comment here for the record - we already agreed that this solution is ok for now.

app/models.js
xoen: 😇 ...I forgot to run `npm lint`

xoen:
> Is this because the test times out at a shorter interval by default?

Correct.

xoen:
(rebased)

xoen:
I pushed some linting commits @andyhd

xoen:
...also bear in mind this depends on https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/63 @andyhd

app/models.js
```diff
@@ -183,17 +183,17 @@ class Pod extends K8sModel {
     // based on
     // https://github.com/kubernetes/dashboard/blob/91c54261c6a3d7f601c67a2ccfbbe79f3b6a89f9/src/app/frontend/pod/list/card_component.js#L98
 
-    let containerStatus = this.data.status.containerStatuses;
+    const containerStatus = this.data.status.containerStatuses;
 
     if (containerStatus) {
-      let state = containerStatus[0].state;
+      const { state } = containerStatus[0];
```
andyhd: 😊  oops

xoen: No problem, we now we're not 100% linted and that we'll do it "lazily" (as in gradually) :)

PR #85, opened 2017-12-14
Ui copy update

axemonkey:
## What

Update copy in the Control Panel based on chat between Robin and Clive ages back.

[Trello](https://trello.com/c/pVJo4ADD/513-2-outcome-of-robin-and-clive-to-iterate-wording-in-ui-to-make-clearer)

## How to review

1. Um. Look at the changes, see that the copy in the app is clearer.

### DIFF
--- a/app/buckets/handlers.js
+++ b/app/buckets/handlers.js
@@ -42,7 +42,8 @@ exports.bucket_details = (req, res, next) => {
         bucket,
         apps_options: apps.exclude(bucket.apps),
         users_options: users.exclude(bucket.users),
-        current_user_is_bucket_admin
+        signedInUserId: req.user.auth0_id,
+        current_user_is_bucket_admin,
       });
     })
     .catch(next);
--- a/app/buckets/handlers.js
+++ b/app/buckets/handlers.js
@@ -42,7 +42,7 @@ exports.bucket_details = (req, res, next) => {
         bucket,
         apps_options: apps.exclude(bucket.apps),
         users_options: users.exclude(bucket.users),
-        signedInUserId: req.user.auth0_id,
+        current_user_id: req.user.auth0_id,
         current_user_is_bucket_admin,
       });
     })
--- a/app/templates/apps/details.html
+++ b/app/templates/apps/details.html
@@ -111,7 +111,7 @@
               <form action="{{ url_for('apps3buckets.delete', {id: apps3bucket.id}) }}" method="post" class="inline-form clearfix">
                 <input type="hidden" name="redirect_to" value="{{ url_for('apps.details', {id: app.id}) }}">
 
-                <input type="submit" class="js-confirm button button-secondary right" value="Revoke access" />
+                <input type="submit" class="js-confirm button button-secondary right" value="Disconnect data source" />
               </form>
             {% endif %}
           </td>
--- a/app/templates/buckets/details.html
+++ b/app/templates/buckets/details.html
@@ -1,14 +1,12 @@
 {% extends "layouts/one-column.html" %}
 
-{% set page_title = "Bucket: " + bucket.name %}
+{% set page_title = "App data source: " + bucket.name %}
 
 {% block main_column %}
-<!-- TODO: This is only a placeholder template - Update me -->
 
 
-<!-- TODO: Placeholder list - double check wording and everything -->
-<h2 class="heading-medium">Users with access</h2>
-<p>{{ bucket.users3buckets.length }} user{%- if bucket.users3buckets.length != 1 -%}s have{% else %} has{% endif %} access to this data source</p>
+<h2 class="heading-medium">Users with admin access</h2>
+<p>{{ bucket.users3buckets.length }} user{%- if bucket.users3buckets.length != 1 -%}s are{% else %} is{% endif %} admin of this app data source</p>
 
 {% if bucket.users3buckets.length %}
   <table class="form-group">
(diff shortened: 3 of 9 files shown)


app/templates/buckets/list.html
```diff
@@ -21,10 +21,12 @@
           <td>{{ bucket.id }}</td>
           <td><a href="{{ url_for('buckets.details', {id: bucket.id}) }}">{{ bucket.name }}</a></td>
           <td class="align-right">
-            <form action="{{ url_for('buckets.delete', {id: bucket.id}) }}" method="post" class="inline-form">
-              <input type="hidden" id="redirect" name="redirect" value="{{ url_for('buckets.list') }}">
-              <input type="submit" class="button button-secondary right js-confirm" value="Delete bucket" data-confirm-message="Are you sure you want to delete this bucket?">
-            </form>
+            {% if current_user_is_bucket_admin or current_user.is_superuser %}
```
xoen: Maybe it's a stupid question but what is `current_user_is_bucket_admin` in the context of the list of apps here?

app/templates/users/includes/tables.html
```diff
@@ -80,10 +80,10 @@ <h2 class="heading-medium">{{ heading_prefix }} app data sources</h2>
           </td>
           <td class="align-right">
             {% if users3bucket.is_admin %}
-              <a class="button button-secondary" href="{{ url_for('buckets.details', {id: users3bucket.id}) }}">Manage bucket access</a>
+              <a class="button button-secondary" href="{{ url_for('buckets.details', {id: users3bucket.id}) }}">Manage data source access</a>
```
xoen: Are we replacing all current instances of the `bucket` word with `app data source`?

app/buckets/handlers.js
```diff
@@ -42,7 +42,8 @@ exports.bucket_details = (req, res, next) => {
         bucket,
         apps_options: apps.exclude(bucket.apps),
         users_options: users.exclude(bucket.users),
-        current_user_is_bucket_admin
+        signedInUserId: req.user.auth0_id,
```
xoen: I wonder if we should call this variable `current_user_id` for consistency with similar variables (with the snake case like the other surrounding variables)

app/templates/apps/details.html
```diff
@@ -111,7 +111,7 @@ <h2 class="heading-medium">App data sources</h2>
               <form action="{{ url_for('apps3buckets.delete', {id: apps3bucket.id}) }}" method="post" class="inline-form clearfix">
                 <input type="hidden" name="redirect_to" value="{{ url_for('apps.details', {id: app.id}) }}">
 
-                <input type="submit" class="js-confirm button button-secondary right" value="Revoke access" />
+                <input type="submit" class="js-confirm button button-secondary right" value="Disconnect data source" />
```
xoen: Is "app" implicit from the context or should this be `Disconnect app data source` (these sentences are getting a bit long...)

app/buckets/handlers.js
andyhd: You don't need to pass this to the template at all - it is already available as `current_user.auth0_id`

app/templates/apps/details.html
axemonkey: In this case yeah, it should be implicit and I left it out to keep the button a bit shorter.

app/buckets/handlers.js
axemonkey: 👍

app/templates/users/includes/tables.html
axemonkey: That's the broad idea, at least in copy that the user sees.

app/templates/buckets/list.html
axemonkey: This is the list of buckets. `current_user_is_bucket_admin` is determining if the logged in user has the right to manage or delete each bucket.

app/buckets/handlers.js
```diff
@@ -42,7 +42,8 @@ exports.bucket_details = (req, res, next) => {
         bucket,
         apps_options: apps.exclude(bucket.apps),
         users_options: users.exclude(bucket.users),
-        current_user_is_bucket_admin
+        current_user_id: req.user.auth0_id,
```
andyhd: You don't need to pass this to the template, it's already available as `current_user.auth0_id`

app/templates/buckets/details.html
```diff
@@ -1,45 +1,49 @@
...
+            <span class="visuallyhidden">Actions</span>
+          {% endif %}
+        </th>
       </tr>
     </thead>
     <tbody>
       {% for users3bucket in bucket.users3buckets %}
         <tr>
-          <td><a href="{{ url_for('users.details', {id: users3bucket.user.auth0_id}) }}">{{ users3bucket.user.username }}</a></td>
+          <td><a class="{% if current_user_id == users3bucket.user.auth0_id %}highlight-current{% endif %}" href="{{ url_for('users.details', {id: users3bucket.user.auth0_id}) }}">{{ users3bucket.user.name }}</a></td>
```
andyhd: You might be able to change this to `if current_user == users3bucket.user`

app/buckets/handlers.js
```diff
@@ -42,7 +42,8 @@ exports.bucket_details = (req, res, next) => {
...
+        current_user_is_bucket_admin,
```
andyhd: You could replace `current_user_is_bucket_admin` in the template with `bucket.has_admin(current_user)`, rather than assigning a variable and passing it.

Or if you add `is_admin(bucket)` to the `User` model, you could  `current_user.is_admin(bucket)`.

app/templates/buckets/list.html
xoen: Isn't being an admin of a bucket something depending on the specific bucket? As far as I can tell this flag doesn't change at each flag iteration.

(Not a massive deal anyway as permissions would be enforced in the API regardless)

app/buckets/handlers.js
xoen: I completely forgot about this `Bucket#has_admin` method 😇

app/buckets/handlers.js
axemonkey: 👍

app/buckets/handlers.js
axemonkey: 👍

app/templates/buckets/details.html
axemonkey: Is that better?

axemonkey: Also, that doesn't work as `users3bucket.user` is a vastly reduced set of data.

andyhd: 👍

PR #88, opened 2017-12-15
Tabbed homepage

axemonkey:
## What

Creates a tabbed interface on the homepage with panes for Tools, Data and Apps preceded by a Superuser tab for superusers.

The Tools tab currently consists of a button which links to `/tools` for now. This can be replaced with the tools view after future work.

## How to review

1. Just review it. I didn't get where I am today by telling you how to review it.
2. Review it.
3. When you've finished reviewing it, stop reviewing it.

### DIFF
--- a/app/assets/javascripts/modules/tabs.js
+++ b/app/assets/javascripts/modules/tabs.js
@@ -1,0 +1,76 @@
+'use strict';
+
+moj.Modules.tabs = {
+  listSelector: 'ul.tabs',
+  panelSelector: 'section.tab-panel',
+  activeClass: 'active',
+
+  init: function() {
+    var self = this;
+
+    self.$list = $(self.listSelector);
+
+    if(self.$list.length) {
+      self.$tabs = self.$list.find('li');
+      self.$panels = $(self.panelSelector);
+
+      self.storeSlugs();
+      self.showInitialTab();
+      self.bindEvents();
+    }
+  },
+
+  bindEvents: function() {
+    var self = this;
+
+    self.$list.on('click', function(e) {
+      self.selectTab($(e.target));
+    });
+  },
+
+  showInitialTab: function() {
+    var self = this;
+    var docHash = document.location.hash;
+    var tabIndex = 0;
+
+    if(docHash.length) {
+      var checkSlug = docHash.slice(1);
+      var slugIndex = self.slugs.indexOf(checkSlug);
+
+      if (slugIndex > 0) {
+        tabIndex = slugIndex;
+      }
+    }
+
+    self.showTab(tabIndex);
+  },
+
+  storeSlugs: function() {
+    var self = this;
+
+    self.slugs = [];
+    self.$tabs.each(function(x, tab) {
+      var tabSlug = encodeURIComponent($(tab).text());
+
+      self.slugs.push(tabSlug);
+    });
+  },
+
+  selectTab: function($tab) {
+    var self = this;
+    var index = self.$tabs.index($tab);
+
+    self.showTab(index);
+  },
+
+  showTab: function(index) {
+    var self = this;
+
+    self.$tabs.removeClass(self.activeClass);
+    self.$panels.removeClass(self.activeClass);
+    self.$tabs.eq(index).addClass(self.activeClass);
+    self.$panels.eq(index).addClass(self.activeClass);
+
+    document.location.hash = self.slugs[index];
+  }
+};
--- a/app/templates/base/home.html
+++ b/app/templates/base/home.html
@@ -5,16 +5,121 @@
 
 {% block main_column %}
 
+  <ul class="tabs">
+    {% if current_user.is_superuser %}
+      <li>Superuser</li>
+    {% endif %}
+    <li>Analytical tools</li>
+    <li>Data</li>
+    <li>Webapps</li>
+  </ul>
+
   {% if current_user.is_superuser %}
-    <h2 class="heading-medium">Superuser functions</h2>
-    <ul class="list list-bullet">
-      <li><a href="{{ url_for('users.list') }}">List all users</a></li>
-      <li><a href="{{ url_for('apps.list') }}">List all apps</a></li>
-      <li><a href="{{ url_for('buckets.list') }}">List all app data sources</a></li>
-    </ul>
+    <section class="tab-panel">
+      <h2 class="heading-medium">Superuser functions</h2>
+      <ul class="list list-bullet">
+        <li><a href="{{ url_for('users.list') }}">List all users</a></li>
+        <li><a href="{{ url_for('apps.list') }}">List all apps</a></li>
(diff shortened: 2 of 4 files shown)


app/templates/base/home.html
```diff
@@ -5,16 +5,133 @@
...
+            </tr>
+          </thead>
+          <tbody>
+            {% for userapp in user.userapps %}
+              <tr>
+                <td>
+                  <a href="{{ url_for('apps.details', {id: userapp.app.id}) }}">{{ userapp.app.name }}</a>
+                </td>
+                <td>
+                  {% if userapp.is_admin %}
```
xoen: You may want to use `{{ yes_no(userapp.is_admin) }}`.

app/templates/base/home.html
```diff
@@ -5,16 +5,133 @@
...
+              <td><a href="{{ url_for('buckets.details', {id: users3bucket.id}) }}">{{ users3bucket.s3bucket.name }}</a></td>
+              <td>
+                {% if users3bucket.is_admin %}
+                  Yes
+                {% else %}
+                  No
+                {% endif %}
+              </td>
+              <td>
+                {% if users3bucket.access_level == 'readonly' %}
```
xoen: You may want to use `{{ yes_no(users3bucket.access_level, 'readwrite') }}`.

app/assets/javascripts/modules/tabs.js
```diff
@@ -0,0 +1,75 @@
...
+      }
+    }
+
+    self.showTab(tabIndex);
+  },
+
+  storeSlugs: function() {
+    var self = this;
+
+    self.slugs = [];
```
xoen: Don't know what's your preference but you may replace this loop with ES6's `map()`:

```javascript
  storeSlugs: function() {
    this.slugs = this.$tabs.map((_, tab) => {
      encodeURIComponent($(tab).text());
    });
  },
```

app/assets/javascripts/modules/tabs.js
```diff
@@ -0,0 +1,75 @@
...
+
+  bindEvents: function() {
+    var self = this;
+
+    self.$list.on('click', function(e) {
+      self.selectTab($(e.target));
+    });
+  },
+
+  showInitialTab: function() {
```
xoen: I get what this is doing and it seems to mostly work as well, but when I refresh a page it doesn't go to the right tab as expected (despite the URL having the hash component)

app/assets/javascripts/modules/tabs.js
axemonkey: This is client-side and we're not transpiling atmo. If `map()` works in ES5, then fine. 😄

app/assets/javascripts/modules/tabs.js
axemonkey: Hmm, odd. Are you using some crazy minority browser like Safari?

app/templates/base/home.html
axemonkey: 👍

app/templates/base/home.html
axemonkey: 👍

app/templates/base/home.html
```diff
@@ -5,16 +5,133 @@
 
 {% block main_column %}
 
+  <ul class="tabs">
```
xoen: So essentially this is an aggregations view of pretty much all the other views.
What does it mean for existing views and redirects?

app/assets/javascripts/modules/tabs.js
xoen: Nope, ignore me, that was me playing with your code and breaking it, sorry Clive 😭

app/assets/javascripts/modules/tabs.js
xoen: It's been a long day...

app/templates/base/home.html
axemonkey: Yep, this is a fair point. The user's homepage aggregates their apps, their data, their tools etc. The only one of these that has an equivalent page at the moment is tools, since the list pages for the others are superuser "show all" views.

Possibly what we could do is if a non-superuser manually enters `/apps` we could redirect them to `/#Webapps` in order to show just their apps, etc.

app/assets/javascripts/modules/tabs.js
axemonkey: 👍

app/assets/javascripts/modules/tabs.js
axemonkey: 👍 🛌

app/assets/javascripts/modules/tabs.js
```diff
@@ -0,0 +1,75 @@
...
+    var self = this,
+      docHash = document.location.hash,
+      tabIndex = 0;
+
+    if(docHash.length) {
+      var checkSlug = docHash.slice(1);
+      var slugIndex = self.slugs.indexOf(checkSlug);
```
andyhd: Inconsistent variable declarations. Please use [one `var` per variable](https://github.com/airbnb/javascript#variables--one-const).

app/assets/javascripts/modules/tabs.js
```diff
@@ -0,0 +1,75 @@
...
+
+    self.showTab(tabIndex);
+  },
+
+  storeSlugs: function() {
+    var self = this;
+
+    self.slugs = [];
+    self.$tabs.each(function(x, tab) {
+      var tabSlug = encodeURIComponent($(tab).text());
```
andyhd: This seems brittle - maybe add a `data-` attribute to the tabs for the slug and use that instead of the text?

app/assets/javascripts/modules/tabs.js
axemonkey: Doh.

app/assets/javascripts/modules/tabs.js
```diff
@@ -34,8 +34,9 @@ moj.Modules.tabs = {
       tabIndex = 0;
 
     if(docHash.length) {
-      var checkSlug = docHash.slice(1);
-      var slugIndex = self.slugs.indexOf(checkSlug);
+      var checkSlug = docHash.slice(1),
+        slugIndex = self.slugs.indexOf(checkSlug);
```
andyhd: No, wrong way: https://github.com/airbnb/javascript#variables--one-const

app/assets/javascripts/modules/tabs.js
axemonkey: I did it this way to keep the HTML completely clean and mean that no extra attributes were needed. Why would this be more brittle than using `data-` attributes?

andyhd: Actually, ignore me - I thought if the tab text changed it would break, but it won't

axemonkey: 👍

PR #89, opened 2017-12-15
Fetch Github repos from Github API

andyhd:
## What

* Added Github API client and model classes
* Fetch Github repos from create app handler (with user's Github access token) and pass to template
* Update create app form to show a dropdown of available Github repos (including private)

## How to review

1. Go to http://localhost:3000/apps/new
2. See the "Github repository URL" form field group has a second dropdown of repo names

See [Trello #509](https://trello.com/c/k4rsqehc)


andyhd:
@axemonkey pointed out that this is not quite working the way we want, so closing for now

app/apps/handlers.js
```diff
@@ -1,12 +1,22 @@
...
-const { repos } = require('../config');
+const { Repo } = require('../models/github');
+const config = require('../config');
+const github = require('../api_clients/github');
+
 
 exports.new = (req, res, next) => {
-  Bucket.list()
-    .then((buckets) => {
+  github.get_access_token(req.user)
```
xoen: I wonder if we could hide this inside the `github` API client and just call `github.authenticate(user)` which would get the access token and authenticate internally, e.g.

```javascript
github.authenticate(user).
    .then(() => {
      return Promise.all([Repo.list(), Bucket.list()]);
    })
    .then(([repos, buckets]) => {
    // ...
```

Or something like that, but I may miss some reason why you added `github.get_access_token()` in the first place.

app/config.js
```diff
@@ -29,6 +29,7 @@ config.apps = [
 config.auth0 = {
   domain: process.env.AUTH0_DOMAIN,
   clientID: process.env.AUTH0_CLIENT_ID,
+  clientId: process.env.AUTH0_CLIENT_ID,
```
xoen: Why do you have both `clientID` and `clientId` with the same value here? I'd just stick with `clientID`.

app/apps/handlers.js
andyhd: That is a great idea 👍

app/config.js
andyhd: Different modules require different capitalisation :(

I'll change it to be more explicit about why

PR #91, opened 2017-12-19
Fetch GitHub repos from the Github API

andyhd:
## What

* Added Github API client and model classes
* Fetch Github repos from create app handler (with user's Github access token) and pass to template
* Update create app form to show a dropdown of available Github repos (including private)

## How to review

1. Go to http://localhost:3000/apps/new
2. See the "Github repository URL" form field group has a second dropdown of repo names

See [Trello #509](https://trello.com/c/k4rsqehc)


app/api_clients/github.js
```diff
@@ -0,0 +1,42 @@
+const config = require('../config');
+const GithubAPI = require('github');
+const { ManagementClient } = require('auth0');
```
xoen: I was wondering if we should keep the same convention we have in python with the order/grouping of requires.

Just an idea but it could help avoid confusion when we have 3rd party library vs. app-specific modules with similar names (e.g. `github` here is not one of our module, to not confuse with things like our `api_clients/github` module).

Obviously **no need to change anything**.

andyhd: Yeah, we should

app/models/github.js
```diff
@@ -0,0 +1,34 @@
...
+          org,
+          type: 'all',
+          page: params.page || 1,
+          per_page: params.per_page || 500
+        });
+      }))
+      .then((results) => {
+        return new ModelSet(
+          this.prototype.constructor,
+          [].concat(...results.map(result => result.data))
```
xoen: I may miss something here so apologies if it's silly.

Isn't this `[].concat()` equivalent to `results.map(result => result.data)` (which should map to an array, unless results is `null` or something like that).

andyhd: `results` is an array of objects with a `data` array property, eg: `[{ data: [1, 2] }, { data: [3, 4] }]`.

I do `results.map(result => result.data)`, which gives me an array of arrays, eg: `[ [1, 2], [3, 4] ]`.

Then I do `[].concat(...<array of arrays>)`, which concats them all into one array, eg: `[1, 2, 3, 4]`.

xoen: 👍

PR #92, opened 2017-12-19
Use ES6 for client-side JS, transpile for browser

axemonkey:
## What

Linting code is good for consistency and quality. To use the same linting engine and rules for the server-side JS and the client-side JS, they must both be written in the same dialect, i.e. ES6. This PR comprises tweaking the client-side JS to ES6 syntax and implements Babel to transpile this into ES5 code for browsers to consume. The linter can then run on the pre-transpiled ES6 code.

## How to review

1. check out and rebuild the front-end so that `npm install` is run
2. use the site normally – nothing should be apparently different
3. run `npm run lint` to see that no errors are reported from client-side code (the server-side code at time of writing reports over 400 problems – to lint just the client-side code go to `package.json` and temporarily change the lint command to point at `app/assets/javascripts` rather than `app`)

TODO: lint the server side JS and implement fixes

### DIFF
--- a/app/assets.js
+++ b/app/assets.js
@@ -4,9 +4,11 @@ const glob = Bluebird.promisify(require('glob').glob);
 const render = Bluebird.promisify(require('node-sass').render);
 const concatfiles = require('concat-files');
 const path = require('path');
+const babel = require('babel-core');
 
 const config = require('./config');
 const log = require('bole')('assets');
+
 
 
 exports.compile_sass = () => {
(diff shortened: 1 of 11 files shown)


app/assets.js
```diff
@@ -39,7 +41,28 @@ function concat_js(files) {
...
+
+
+function transpile_js() {
+  let inFile = path.join(config.js.outDir, config.js.filename);
+  let outFile = path.join(config.js.outDir, 'transpiled_' + config.js.filename);
+
+  log.info(
+    `transpiling ${relative(inFile)} -> ${relative(outFile)}`);
+
+  babel.transformFile(inFile, { "presets": ["es2015"] }, function(err, result) {
```
andyhd: Would it make sense to get this `{"presets": ["es2015"]}` from `config.js`?

axemonkey: Yes, good call. Will do that.

xoen:
Nice 👍

PR #93, opened 2017-12-20
Show tools in tabbed home page

xoen:
### What
The main aim of this is to display the list of tools directly in the new tabbed home without having to click around.

#### Other UX improvements include
 - Added "Open" button for tools as the link wasn't explicit enough for the users
 - Don't ask for confirmation when deploying RStudio. It's not a destructive operation so no point in asking.
 - Support for redirect to specific tab of new homepage
 - Removed `GET /tools` view

#### Internal refactoring
 - Added `Tool` model with `Tool.list` to not have logic to get k8s deployments/pods directly in route handler
 - Added support for [fragments](https://en.wikipedia.org/wiki/Fragment_identifier) (the part after the `#` in an URL) in `routes.url_for()`
   - All existing client code is updated but worth watch out for anything I missed
 - Removed `GET /tools` route/handler
 - Moved tools' `list.html` under `/app/templates/includes/list.html` for use in other views (for now it's only used in the home but we may want to render the list of tools in other places)
 - Moved `get_tool_url()` in `tools/helpers` module instead of having it directly in the route handler code
 - Don't make unnecessary request to get current user in home route handler and don't pass unnecessary locals, just use `current_user` which is already available.
    - For example `signedInUser` was passed always as `true` so I removed few unnecessary `if` statements
 - More linting love

### Notable changes

In order to support fragment I had to change the function signature and update all the client code.

Before:
`url_for('buckets.details', { id: bucket_id })`

Now:
`url_for('buckets.details', { params:  { id: bucket_id } })`

To get the URL with the fragment part you can pass the `fragment` argument:
`url_for('base.home', { fragment: 'Analytical tools' })`

### How to review

1. Start the thing
2. Click around
3. Check things are working as expected
4. Sell some bitcoins
5. Profit!

### Ticket
https://trello.com/c/rp8rfsWy/568-3-refactor-for-tools-tab-in-homepage

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -33,23 +33,25 @@ function add_route(route, router) {
   router[route.method](route.pattern, route.handler);
 }
 
-
-exports.url_for = function (route_name, args = {}) {
-
-  let route = routes[route_name];
+exports.url_for = function (route_name, {params = {}, fragment} = {}) {
+  const route = routes[route_name];
 
   if (!route) {
     throw new Error(`route ${route_name} not found`);
   }
 
   try {
-    return replace_route_params(route.pattern, args) + query_string(args);
+    let url = replace_route_params(route.pattern, params) + query_string(params);
 
+    if (fragment) {
+      url += '#' + encodeURIComponent(fragment);
+    }
+
+    return url;
   } catch (error) {
     throw new Error(`url_for("${route_name}") failed: ${error}`);
   }
 };
-
 
 function replace_route_params(pattern, param_values) {
 
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -53,7 +53,7 @@ exports.create = (req, res, next) => {
         })
         .then(() => {
           const { url_for } = require('../routes'); // eslint-disable-line global-require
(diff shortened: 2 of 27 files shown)


app/routes.js
```diff
@@ -33,24 +33,26 @@ function add_route(route, router) {
   router[route.method](route.pattern, route.handler);
 }
 
-
-exports.url_for = function (route_name, args = {}) {
-
-  let route = routes[route_name];
+exports.url_for = function (route_name, {params = {}, fragment} = {}) {
```
andyhd: On second thought, I think this change should be reverted. It makes `url_for` less easy to use and we only use fragments in 3 places. Instead use `${url_for('handler.name', { id: ... })}#${escape(fragment)}`.

xoen: 😱  - Actually makes sense, I'll have a look.

xoen: FYI: I tried to move commits around and/or revert that commit but it's been a pain so it seems I'll go for the simple option of merging this and the other PR and "revert" this `url_for()` changes manually 😞

xoen: (as discussed offline few minutes ago)

PR #98, opened 2017-12-22
Remove support for `url_for()` fragment argument

xoen:
(Manual revert of commit 24169260357d26bafe02061d2425dcfc289dc766)

See: https://github.com/ministryofjustice/analytics-platform-control-panel-frontend/pull/93#discussion_r158285305


xoen:
I searched again throughout the all codebase and it seems like I didn't forget to change of the `url_for()` calls, but feel free to double check if your want @andyhd

PR #102, opened 2018-01-04
get User fresh when homepage loads

axemonkey:
## What

`current_user` doesn't seem to get updated when adding an app or bucket for the first time

To reproduce:

1. check out master, **wipe DB**, create (super)user
2. create new app and/or new data source
3. go to `/` and click 'Data' or 'Webapps'
4. see that your newly created items are not shown

## How to review

1. check out branch
2. repeat steps above
3. see that your newly created items are shown on the relevant homepage tabs

### DIFF
--- a/app/base/handlers.js
+++ b/app/base/handlers.js
@@ -12,9 +12,14 @@ exports.home = (req, res, next) => {
   const rstudio_is_deploying = req.session.rstudio_is_deploying;
   req.session.rstudio_is_deploying = false;
 
-  Tool.list()
-    .then((tools) => {
-      res.render('base/home.html', { tools, get_tool_url, rstudio_is_deploying });
+  Promise.all([Tool.list(), User.get(req.user.auth0_id)])
+    .then(([tools, user]) => {
+      res.render('base/home.html', {
+        tools,
+        get_tool_url,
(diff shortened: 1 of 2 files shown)


app/base/handlers.js
```diff
@@ -12,9 +12,14 @@ exports.home = (req, res, next) => {
   const rstudio_is_deploying = req.session.rstudio_is_deploying;
   req.session.rstudio_is_deploying = false;
 
-  Tool.list()
-    .then((tools) => {
-      res.render('base/home.html', { tools, get_tool_url, rstudio_is_deploying });
+  Promise.all([Tool.list(), User.get(req.user.auth0_id)])
+    .then(([tools, user]) => {
```
andyhd: Might be good to update `req.user` here with `user`

andyhd: Then you won't have to pass `user` to the template because `current_user` will be up to date, and then you don't need the template changes.

axemonkey: I don't follow... would updating `req.user` change `app.locals.current_user`? I tried simply assigning it to `user` and nothing changed...

axemonkey: Leaving as is, as discussed. 👍

PR #104, opened 2018-01-05
replace npm with yarn

axemonkey:
## What

Yarn is nicer than npm. Also, it's deterministic and pegs everything to a version with the `yarn.lock` file.

## How to review

1. `brew install yarn`
2. Check out branch, do a local rebuild. Everything should work as expected.
3. The README has updated commands for running, testing and linting. Try those.


xoen:
Nice one 👍 - I'm liking yarn so far from the very little I've seen.

I'll need to wrap my head around how the locking works in practice and whether we need to change anything in the `Dockerfile`, etc...

PR #109, opened 2018-01-10
Repo typeahead

axemonkey:
## What

When adding a new app to link to some data, it is necessary to link an existing repo from one of the allowed github orgs. This means a long dropdown with ~100 entries at the moment. Implementing a typeahead should make it easier to find a repo from the specified pool.

## How to review

1. check out, rebuild, run
2. create a new app
3. see that where there was previously a dropdown listing repos, there is now a typeahead search box.

Before
---
image

After
---
image

### DIFF
--- a/app/assets/javascripts/modules/repo-description.js
+++ b/app/assets/javascripts/modules/repo-description.js
@@ -5,7 +5,7 @@ moj.Modules.repoDescription = {
   repoSlugTypeaheadName: 'repo_typeahead',
   repoUrlHiddenInputName: 'repo_url',
   descriptionInputName: 'description',
-  repoSelected: null,
+  repoSelected: false,
 
   init() {
     this.$repoSlugSelect = $(`#${this.repoSlugSelectName}`);
@@ -31,14 +31,8 @@ moj.Modules.repoDescription = {
     const repos = repoOptions.map(opt => opt.text);
     const descriptions = repoOptions.map(opt => opt.dataset.description);
 
-    // reset potentially pre-existing typeahead
-    this.repoSelected = false;
-    this.$repoSlugSelect.hide();
-    this.$repoSlugTypeahead.val('');
-    $('.typeahead__result, .typeahead__cancel-button').remove();
-    $('.typeahead__container').removeClass('cancel');
+    this.resetTypeahead();
 
-    // initialise repo-typeahead for selected org
     this.$repoSlugTypeahead.typeahead({
       order: 'asc',
       maxItem: 0,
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/modules/repo-description.js
```diff
@@ -18,32 +20,58 @@ moj.Modules.repoDescription = {
...
     });
   },
 
   showOrgRepos() {
     const currentOrg = this.$orgSelect.val();
+    const repoOptions = Array.from($(`optgroup[label="${currentOrg}"] option`));
+    const repos = repoOptions.map(opt => opt.text);
+    const descriptions = repoOptions.map(opt => opt.dataset.description);
+
+    // reset potentially pre-existing typeahead
```
xoen: You may potentially move this in a function, something like `resetTypeahead()` which resets the typeahead component to its original state.

app/assets/javascripts/modules/repo-description.js
```diff
@@ -18,32 +20,58 @@ moj.Modules.repoDescription = {
...
+        data: repos,
+      },
+      callback: {
+        onClickAfter: (node) => {
+          const index = repos.indexOf(node[0].value);
+
+          this.getRepoDescription(descriptions[index]);
+          this.repoSelected = true;
+        },
+        onCancel: () => {
```
xoen: Is this block conceptually doing the same thing done when you reset the typeahead component?

app/templates/apps/new.html
```diff
@@ -28,6 +28,15 @@
       {% endfor %}
     </select>
     <span>/</span>
+
+    <div class="typeahead__container">
+      <div class="typeahead__field">
+        <span class="typeahead__query">
```
xoen: Is this span `. typeahead__query` necessary, maybe it can be removed and the `<input>` could go directly under `div.typeahead__field` or maybe even have the `__field` class.

Not a problem obviously, I was just curious about this structure :)

app/assets/javascripts/modules/repo-description.js
andyhd: I agree - wherever you have a comment, you can probably refactor into a self-documenting function

app/assets/javascripts/modules/repo-description.js
```diff
@@ -18,32 +20,58 @@ moj.Modules.repoDescription = {
...
+          this.updateDescription('');
+          $('#repo-results').addClass('js-hidden');
+          this.repoSelected = false;
+        },
+        onSubmit: () => {
+          if (!this.repoSelected) {
+            return false;
+          }
+          return true;
```
andyhd: Looks like this could be replaced with `return this.repoSelected;`

andyhd: I think this function prevents submitting unless a repo is selected, right? Perhaps this could be more explicit - do event handlers still take an event object?, so you can do something like:
```javascript
onSubmit: (event) => {
  if (!this.repoSelected) {
    event.preventDefault();
  }
}
```

axemonkey: `preventDefault()` doesn't work in this case, I assume because it is attached to the typeahead rather than the form's submit event directly.

The shorthand in your previous comment is way better though. 👍

app/templates/apps/new.html
axemonkey: Very probably, but the default configuration [expects a certain HTML structure](http://www.runningcoder.org/jquerytypeahead/documentation/), and it didn't seem worth the effort to streamline it.

app/assets/javascripts/modules/repo-description.js
axemonkey: 👍

app/assets/javascripts/modules/repo-description.js
axemonkey: 👍

app/templates/apps/new.html
xoen: Uh, didn't find that in my 2 minutes search earlier (just had a very quick look at the npm module documentation). As I said, not an issue, I was just curious ;)

PR #110, opened 2018-01-12
Tidy up

andyhd:
## What

* Tidied up some old tests and added some new
* Nicely format Django error messages from the API
* Show request headers and body in the Debug panel on the error page
* Refactor `yes_no` into a template macro

## How to review

1. Go to http://localhost:3000/error?forbidden=1 and see request metadata in the Debug panel
2. Somehow get an API error and see a nicely formatting Django stacktrace on the error page


xoen:
I see you moved the k8s tests under `tests/models`. I also see a bunch of deleted tests but I can't see them being moved to files under `tests/models`, am I missing something? Maybe they need to be git-added?

app/api_clients/control_panel_api.js
```diff
@@ -11,10 +11,55 @@ class ControlPanelAPIClient extends APIClient {
...
 }
 
 exports.APIClient = ControlPanelAPIClient;
 
 const api = new ControlPanelAPIClient(config.api);
 
 exports.api = api;
+
+
+class DjangoError extends APIError {
```
xoen: I wonder if we could keep all these errors classes in a `app/api_clients/errors.js`.

andyhd:
The bucket and user model tests are redundant - covered by test/models/control_panel_api.js

app/api_clients/control_panel_api.js
andyhd: The DjangoError is pretty specific to the control_panel_api, so I think it belongs with it

test/test-tool-model.js
```diff
@@ -1,36 +0,0 @@
-const { assert } = require('chai');
-const { mock_api } = require('./conftest');
-const { api } = require('../app/api_clients/kubernetes.js');
-const { Deployment, ModelSet, Pod, Tool } = require('../app/models');
-
-
-describe('Tool model', () => {
```
xoen: Are we happy to get rid of this test?

andyhd: I moved this into test/models/kubernetes.js along with some small changes

PR #113, opened 2018-01-16
Open rstudio if running

andyhd:
## What

* Hide "Open" button if RStudio is not running.
* Tweaked routes to allow requiring `url_for` globally
* Fixed other eslint exceptions for global requires
* Tidy up a warning message in assets transpiling code

## How to review

1. Go to your tools list
2. Restart your Rstudio and quickly reload the page
3. You should see no "Open" button while there are no "Running" pods.

### DIFF
--- a/app/assets.js
+++ b/app/assets.js
@@ -27,7 +27,12 @@ function transpile_js() {
       process.exit(1);
     }
 
-    fs.writeFile(outFile, result.code);
+    fs.writeFile(outFile, result.code, (err) => {
+      if (err) {
+        log.error(err);
+        process.exit(1);
+      }
+    });
   });
 }
 
--- a/app/models/kubernetes.js
+++ b/app/models/kubernetes.js
@@ -44,23 +44,25 @@ class Deployment extends Model {
     return 'deployments';
   }
 
+  get_pods() {
+    if (!this.pods.length) {
+      return Pod.list({ labelSelector: `app=${this.app_label}` }).
+        then((pods) => {
+          this._pods = pods;
+          return pods;
+        });
+    }
+    return Promise.resolve(this.pods);
+  }
+
   static list() {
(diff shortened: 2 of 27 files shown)


app/assets.js
```diff
@@ -27,7 +27,12 @@ function transpile_js() {
       process.exit(1);
     }
 
-    fs.writeFile(outFile, result.code);
+    fs.writeFile(outFile, result.code, (err) => {
```
axemonkey: 👍  Thank you. I'd been meaning to come back and figure out what I did wrong.

app/models/kubernetes.js
```diff
@@ -34,33 +35,62 @@ exports.Model = Model;
...
+    return this.data.pods.some(pod => pod.status.phase === 'Running');
+  }
+
+  get app_label() {
+    return this.data.metadata.labels.app;
+  }
+
+  get username() {
+    // might get a truncated username
+    return this.data.metadata.namespace.split('user-').join('');
```
xoen: One minor thing is that if the username starts with `user-`, you may get a namespace like `user-user-42` and this would not work.

We probably don't have any case like that so don't worry too much I guess, it's just another of these edge cases to bear in mind (especially if it comes to security).

app/models/kubernetes.js
```diff
@@ -93,56 +123,32 @@ class Pod extends Model {
...
-        pods.forEach((pod) => {
-          const tool = tools_lookup[pod.metadata.labels.app];
-          if (tool) {
-            tool.pods.push(pod);
-          }
-        });
 
-        return tools;
-      });
+  get display_status() {
```
xoen: I like this separation between "get the status" and "display the status" 👍

app/models/kubernetes.js
andyhd: That is a good point.

PR #117, opened 2018-01-22
Show user to be an admin of the app/bucket they just created

axemonkey:
## What

Values stored under `is_admin` are lowercase but being checked against `True` in the `yesno` macro. Booleans in JS are lowercase, so have changed the default `true` value.


axemonkey:
Ah, fair point about Nunjucks, mea culpa. This only got picked up because Robin was demoing something to me last week and on creating a new app it showed him as not being admin despite being listed in the admins. :)

PR #118, opened 2018-01-22
Refactor API clients for thread/process safety

andyhd:
## What

There was a bug where a user restarted their RStudio instance and got a flash message telling them that another user's instance was being restarted.

The problem may have been caused by a fundamental mistake in that the API client instances were singletons and their authentication state may be shared between requests.

It may have been possible that an API client instance was authenticated with User A's credentials, then in the next request from User B, that same API client instance was used to access the API - still using User A's credentials.

This refactor removes the singleton instances and instead instantiates API clients on every request, and stores them in a namespace object in scope only for a specific request - this is enabled by the `continuation-local-storage` module. This should remove the possibility that an API client instance might be shared between requests.

### DIFF
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -1,12 +1,16 @@
 const { App, Bucket, User } = require('../models');
 const { Repo } = require('../models/github');
+const cls = require('continuation-local-storage');
 const config = require('../config');
-const github = require('../api_clients/github');
+const { GithubAPIClient } = require('../api_clients/github');
 const { url_for } = require('../routes');
 
 
 exports.new = (req, res, next) => {
-  github.api.authenticate(req.user)
+  const github_api = new GithubAPIClient(config);
+  const ns = cls.getNamespace(config.continuation_locals.namespace);
+  ns.set('github', github_api);
+  github_api.authenticate(req.user)
     .then(() => Promise.all([Repo.list(), Bucket.list()]))
     .then(([repos, buckets]) => {
       res.render('apps/new.html', {
--- a/app/config.js
+++ b/app/config.js
@@ -47,6 +47,10 @@ config.cluster = {
   tools_domain: process.env.TOOLS_DOMAIN,
 };
 
+config.continuation_locals = {
+  namespace: 'cpfrontend',
+};
+
 config.ensure_login = {
   exclude: [
     /^\/callback/,
@@ -74,11 +78,12 @@ config.js = {
 config.log = {
   requests: process.env.ENABLE_ACCESS_LOGS !== 'false',
   stream: process.stdout,
-  level: process.env.LOG_LEVEL || 'debug',
+  level: 'debug',
 };
 
 // order is important!
 config.middleware = [
+  'continuation-locals',
   'raven-request-handler',
   'request-logging',
   'static',
--- a/app/config.js
+++ b/app/config.js
@@ -78,7 +78,7 @@ config.js = {
 config.log = {
   requests: process.env.ENABLE_ACCESS_LOGS !== 'false',
   stream: process.stdout,
-  level: 'debug',
+  level: process.env.LOG_LEVEL || 'debug',
 };
 
 // order is important!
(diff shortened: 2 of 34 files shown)


app/config.js
```diff
@@ -73,11 +77,12 @@ config.js = {
 config.log = {
   requests: process.env.ENABLE_ACCESS_LOGS !== 'false',
   stream: process.stdout,
-  level: process.env.LOG_LEVEL || 'debug',
```
axemonkey: ? Removing ability to override log level in env?

andyhd: Oops, didn't mean to commit that. `LOG_LEVEL` is also used by the API and so the settings clash and I get no logging from the frontend, so I was temporarily ignoring the env var. I will remove this.

app/middleware/continuation-locals.js
```diff
@@ -0,0 +1,18 @@
...
+module.exports = (app, conf, log) => {
+  log.info('adding continuation local storage');
+
+  const ns = cls.createNamespace(conf.continuation_locals.namespace);
+
+  return (req, res, next) => {
+    ns.bindEmitter(req);
+    ns.bindEmitter(res);
+
+    ns.run(() => {
```
xoen: Is this test code or needs to be done on each request?

app/apps/handlers.js
```diff
@@ -1,12 +1,16 @@
...
-const github = require('../api_clients/github');
+const { GithubAPIClient } = require('../api_clients/github');
 const { url_for } = require('../routes');
 
 
 exports.new = (req, res, next) => {
-  github.api.authenticate(req.user)
+  const github_api = new GithubAPIClient(config);
+  const ns = cls.getNamespace(config.continuation_locals.namespace);
+  ns.set('github', github_api);
```
xoen: Few questions: Where is the stuff set goes? Is there any limitation on how big the values set in the namespace are? Do you reckon serialisation of these objects would cause problems?

app/middleware/continuation-locals.js
andyhd: This is what associates the context with the function call chain

app/apps/handlers.js
andyhd: The objects are not serialized - the namespace object is in memory for the duration of the request handling function call chain (eg: continuation middleware calls `next()` -> auth middleware calls `next()` -> route handler calls `res.send()`), then it is destroyed.

test/test-login.js
```diff
@@ -35,18 +33,16 @@ describe('Logging in', () => {
         handlers.auth_callback[1](req, res, reject);
       });
 
-      const user_details_request = mock_api()
-        .get(`/users/${escape(user.auth0_id)}/`)
-        .matchHeader('Authorization', `JWT ${id_token}`)
-        .reply(200, user);
+      //const user_details_request = mock_api()
```
xoen: Isn't `mock_api()`/nock working in the context of `withAPI()`?

app/middleware/continuation-locals.js
xoen: Do you need to call `ns.set('test', 'foo')` in order for this to happen? If so it may be worth adding a comment and/or change the key/value to something more revealing.

app/apps/handlers.js
xoen: I had a quick look at the module code and that's what I thought 👍

xoen:
I'm not familiar with this pattern/module but as far as I can tell it looks OK (I asked few of my usual stupid questions)  👍

app/middleware/continuation-locals.js
andyhd: Sorry, didn't see what you were talking about. I wil remove that.

test/test-login.js
andyhd: No - `mock_api`/nock mocks `http.request` and `http.clientRequest` to return a specified response. `withAPI` ensures that the call to `handlers.auth_callback` is bound to the namespace containing the API clients.

PR #119, opened 2018-01-24
Remove login template and Auth0 Lock, switch to hosted login for SSO win

axemonkey:
## What

That ^

## How to review

1. check out, do the docker down/pull/build/up dance probably
2. go to `/`
3. see that you are redirected to the "kubectl-oidc" login popup on Auth0 (titling and styling to come)
4. log in
5. see that you are returned to the control panel as usual


app/base/handlers.js
```diff
@@ -55,14 +55,18 @@ exports.auth_callback = [
...
-  req.session[config.auth0.sessionKey] = { state };
-  res.render('login.html', {
-    env: process.env,
-    session: req.session,
-    state,
-  });
+exports.login = (req, res, next) => {
+  if (req.isAuthenticated()) {
+    if (/^http/.test(req.session.returnTo)) {
+      res.send(400, 'URL must be relative');
```
andyhd: This doesn't look like it will render a nice error page.

axemonkey: Yep, completely forgot about that. 👍

PR #122, opened 2018-01-31
Display error message when creating app with used repo URL

andyhd:
## What

* Refactored `app.create` for readability
* Pass required Github repos list and S3buckets list to create app form on error - this was the cause of the error message: the template expects to have repos and buckets for the form inputs, but the form was being rendered with only the error object.
* Add regression test for this issue
* Pre-populate form with entered repo name

## How to review

1. Login and create an app 
2. Now try to create another app with the same repo URL
3. See a form error message highlighting the URL input

See https://github.com/ministryofjustice/analytics-platform/issues/20

### DIFF
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -19,56 +19,69 @@ exports.new = (req, res, next) => {
         repos,
         bucket_prefix: `${process.env.ENV}-`,
         buckets,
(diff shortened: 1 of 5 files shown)


andyhd:
Broke some other flows, closing.

PR #124, opened 2018-01-31
fix IDs for s3 bucket operations

axemonkey:
## What

Instances of `users3bucket.id` should have been `users3bucket.s3bucket.id`, and links/actions were wrong because of that.\

Fixes https://github.com/ministryofjustice/analytics-platform/issues/22

### DIFF
--- a/app/templates/buckets/details.html
+++ b/app/templates/buckets/details.html
@@ -30,7 +30,7 @@
           <td class="align-right">
             {% if bucket.has_admin(current_user) or current_user.is_superuser %}
               <!-- Button to change access level -->
-              <form action="{{ url_for('users3buckets.update', { id: users3bucket.id }) }}" method="post" class="inline-form clearfix">
+              <form action="{{ url_for('users3buckets.update', { id: users3bucket.s3bucket.id }) }}" method="post" class="inline-form clearfix">
                 <input type="hidden" name="access_level" value="{{ yes_no(users3bucket.access_level, 'readwrite', 'readonly', 'readwrite') }}" />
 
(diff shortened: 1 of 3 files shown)


app/templates/buckets/details.html
```diff
@@ -30,7 +30,7 @@ <h2 class="heading-medium">Users with admin access</h2>
           <td class="align-right">
             {% if bucket.has_admin(current_user) or current_user.is_superuser %}
               <!-- Button to change access level -->
-              <form action="{{ url_for('users3buckets.update', { id: users3bucket.id }) }}" method="post" class="inline-form clearfix">
+              <form action="{{ url_for('users3buckets.update', { id: users3bucket.s3bucket.id }) }}" method="post" class="inline-form clearfix">
```
xoen: Watch out @axemonkey - this was indeed correct 😱 - in this case you're updating an `UserS3Bucket` so the ID needs to be the one of the `UserS3Bucket`, not of the `S3Bucket`.

app/templates/buckets/details.html
```diff
@@ -39,7 +39,7 @@ <h2 class="heading-medium">Users with admin access</h2>
               </form>
 
               <!-- Button to revoke access -->
-              <form action="{{ url_for('users3buckets.delete', { id: users3bucket.id }) }}" method="post" class="inline-form clearfix">
+              <form action="{{ url_for('users3buckets.delete', { id: users3bucket.s3bucket.id }) }}" method="post" class="inline-form clearfix">
```
xoen: See other comment: https://github.com/ministryofjustice/analytics-platform-control-panel-frontend/pull/124/files#r165109230

I think this was correct as you're revoking access, therefore deleting an `UserS3Bucket`, not an `S3Bucket`.

PR #126, opened 2018-02-01
fix logout with returnTo URL

axemonkey:
## What

Logout on Alpha was returning the user to the authentication flow instead of the login page, because it was logging out of the application but not from SSO. This fixes that by hitting the auth0 logout endpoint, and supplying a `returnTo` value.

NB: will break any (non-local-dev) environments it is merged to unless the environment vars `APP_PROTOCOL`, `APP_HOST` and `APP_HOST` are present.

e.g.

```
APP_PROTOCOL=http
APP_HOST=localhost
APP_PORT=3000
```

## How to review

1. check out, docker down/up/in/out/etc
2. log in, log out, log in again
3. see that everything works as it should

### DIFF
--- a/app/base/handlers.js
+++ b/app/base/handlers.js
@@ -67,7 +67,8 @@ exports.logout = (req, res) => {
     }
   });
   req.session.destroy(() => {
+    const encodedReturnTo = encodeURI(`${config.app.protocol}://${config.app.host}:${config.app.port}`);
     res.clearCookie(config.session.name);
-    res.redirect('/');
+    res.redirect(`https://${config.auth0.domain}/v2/logout?returnTo=${encodedReturnTo}&client_id=${config.auth0.clientID}`);
   });
 };
--- a/app/base/handlers.js
+++ b/app/base/handlers.js
@@ -6,6 +6,11 @@ const { url_for } = require('../routes');
 const uuid = require('uuid');
 
 
+function sso_logout_url() {
+  const returnTo = encodeURI(`${config.app.protocol}://${config.app.host}:${config.app.port}`);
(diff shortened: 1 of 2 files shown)


app/config.js
```diff
@@ -13,6 +13,9 @@ config.api = {
 config.app = {
   env: process.env.ENV || 'dev',
   asset_path: '/static/',
+  protocol: process.env.APP_PROTOCOL || 'http',
+  host: process.env.APP_HOST || 'localhost',
+  port: process.env.APP_PORT || '3000',
```
andyhd: Should `port` be `process.env.APP_PORT || process.env.EXPRESS_PORT || 3000` ?

app/base/handlers.js
```diff
@@ -67,7 +67,8 @@ exports.logout = (req, res) => {
     }
   });
   req.session.destroy(() => {
+    const encodedReturnTo = encodeURI(`${config.app.protocol}://${config.app.host}:${config.app.port}`);
     res.clearCookie(config.session.name);
-    res.redirect('/');
+    res.redirect(`https://${config.auth0.domain}/v2/logout?returnTo=${encodedReturnTo}&client_id=${config.auth0.clientID}`);
```
andyhd: I would return this value from a self-documenting function (even though it's only used once).
Also, the `/v2/logout` bit looks like it should be in configuration, eg:
 eg:
```javascript
function sso_logout_url() {
  const returnTo = encodeURI(...);
  return `https://${config.auth0.domain}${config.auth0.sso_logout_url}?returnTo=${returnTo}&client_id=${config.auth0.clientID}`;
}
```

app/config.js
axemonkey: Sure 👍

app/base/handlers.js
axemonkey: 👍

PR #128, opened 2018-02-06
Reworks create app page to allow freely typed repo URLs

axemonkey:
## What

Previously an app could only be created if the URL was created by matching a value in the typeahead repo box. This meant that the user could only create apps matching repos that they have access to in github. It has been requested that this be changed so that an app may be created with any freely typed URL in the typeahead box.

## How to review

1. check you can still create apps using the typeahead as intended, for either org, and that descriptions still update as expected
2. then try creating apps by typing freely in the typeahead box
3. both should work, and the description matching should still work when an existing repo is matched
4. check that the form submit button is disabled when the typeahead box is empty, but enabled when it contains any value

### DIFF
--- a/app/assets/javascripts/modules/repo-description.js
+++ b/app/assets/javascripts/modules/repo-description.js
@@ -1,21 +1,26 @@
 moj.Modules.repoDescription = {
+  formId: 'create_app',
   repoPrefix: 'https://github.com',
   orgSelectName: 'repo_org',
   repoSlugSelectName: 'name',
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/modules/repo-description.js
```diff
@@ -1,62 +1,75 @@
 moj.Modules.repoDescription = {
```
andyhd: This module name doesn't really reflect what it does any more

axemonkey: Very good point.

PR #129, opened 2018-02-13
Enable user delete button

andyhd:
* Make the delete user button actually delete the user.
* Display a note in the flash message about waiting a few seconds for deletion to complete
* Redirect to user list (only superusers should be able to delete users)

### DIFF
--- a/app/templates/users/list.html
+++ b/app/templates/users/list.html
@@ -20,13 +20,15 @@
   <tbody>
     {% for user in users %}
       <tr>
-        <td><a class="{% if current_user.auth0_id == user.auth0_id %}highlight-current{% endif %}"" href="{{ url_for('users.details', { id: user.auth0_id }) }}">{{ user.name | default(user.username, true) }}</a></td>
+        <td><a class="{% if current_user.auth0_id == user.auth0_id %}highlight-current{% endif %}" href="{{ url_for('users.details', { id: user.auth0_id }) }}">{{ user.name | default(user.username, true) }}</a></td>
         <td>{{ user.username }}</td>
         <td>{{ user.email }}</td>
         {% if current_user.is_superuser %}
           <td class="align-right">
             <a class="button button-secondary" href="{{ url_for('users.edit', { id: user.auth0_id }) }}">Edit</a>
-            <a class="button button-secondary js-confirm" href="#">Delete</a>
+            <form action="{{ url_for('users.delete', { id: user.auth0_id }) }}" method="post" class="inline-form" right clearfix">
+              <input type="submit" class="button button-warning right js-config" value="Delete" data-confirm-message="Are you sure you want to delete this user?">
+            </form>
           </td>
         {% endif %}
       </tr>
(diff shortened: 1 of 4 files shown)


app/templates/users/list.html
```diff
@@ -20,13 +20,15 @@
...
       <tr>
-        <td><a class="{% if current_user.auth0_id == user.auth0_id %}highlight-current{% endif %}"" href="{{ url_for('users.details', { id: user.auth0_id }) }}">{{ user.name | default(user.username, true) }}</a></td>
+        <td><a class="{% if current_user.auth0_id == user.auth0_id %}highlight-current{% endif %}" href="{{ url_for('users.details', { id: user.auth0_id }) }}">{{ user.name | default(user.username, true) }}</a></td>
         <td>{{ user.username }}</td>
         <td>{{ user.email }}</td>
         {% if current_user.is_superuser %}
           <td class="align-right">
             <a class="button button-secondary" href="{{ url_for('users.edit', { id: user.auth0_id }) }}">Edit</a>
-            <a class="button button-secondary js-confirm" href="#">Delete</a>
+            <form action="{{ url_for('users.delete', { id: user.auth0_id }) }}" method="post" class="inline-form" right clearfix">
```
axemonkey: Typo - need to lose a " here in the `class`

app/users/routes.js
```diff
@@ -8,4 +8,5 @@ module.exports = [
   { name: 'edit', pattern: '/users/:id/edit', handler: handlers.user_edit },
   { name: 'verify_email', pattern: '/verify-email', handler: handlers.verify_email },
   { name: 'verify_email', method: 'POST', pattern: '/verify-email', handler: handlers.verify_email },
+  { name: 'delete', method: 'POST', pattern: '/delete', handler: handlers.delete },
```
xoen: Does this needs to receive the `id` param? Maybe change the route to `/users/:id/delete`?

app/templates/users/list.html
andyhd: Oops

test/users/test-delete.js
```diff
@@ -0,0 +1,17 @@
+const { assert } = require('chai');
+const { dispatch, mock_api, url_for, user } = require('../conftest');
+const handlers = require('../../app/users/handlers');
+
+
+describe('users', () => {
+  describe('delete', () => {
+    it('deletes user and redirects to list', () => {
+      mock_api().delete(`/users/${escape(user.auth0_id)}/`).reply(200, user);
```
xoen: Not that it matters much in the test I think, but if I remember correctly `DELETE` endpoints returns [`204 No Content`](https://httpstatusdogs.com/204-no-content) on success.

andyhd: Too slow, I already changed that :)

xoen: Sorry if I'm being a pain with these comments. May be a good idea to test that this call to the backend was made as well. But ignore me if it's too much hassle.

xoen: Ah :D

xoen: Don't know why but feeling really slow this afternoon :P

andyhd: Yep, I missed that. Thanks.

PR #131, opened 2018-02-13
Fixed logout in dev/alpha

xoen:
### What
When Auth0 redirect the user back to port `https://{host}:80` this
doesn't work as HTTPS traffic is on port `443`.

Removed `app.port` config (it can be passed as part of `app.host`
for when developing locally).

### How to review

1. Click logout
2. (the Auth0 client needs to have your test URL in the `Allowed Logout URLs` section)
3. You get logged out, without seeing and error

### DIFF
--- a/app/base/handlers.js
+++ b/app/base/handlers.js
@@ -7,7 +7,7 @@ const uuid = require('uuid');
 
 
 function sso_logout_url() {
-  const returnTo = encodeURI(`${config.app.protocol}://${config.app.host}:${config.app.port}`);
+  const returnTo = encodeURI(`${config.app.protocol}://${config.app.host}`);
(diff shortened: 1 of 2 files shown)


app/base/handlers.js
```diff
@@ -7,7 +7,7 @@ const uuid = require('uuid');
 
 
 function sso_logout_url() {
-  const returnTo = encodeURI(`${config.app.protocol}://${config.app.host}:${config.app.port}`);
+  const returnTo = encodeURI(`${config.app.protocol}://${config.app.host}`);
```
andyhd: Should this be removed completely? I think maybe it should be possible to set it to `null` or `undefined` and only add it to the `returnTo` URL if it is set to a number.

andyhd: I think omitting it would break logout when running locally

xoen: As mentioned in PR description, when running the UI locally, just add it to the `APP_HOST`, in fact the default value is `localhost:3000`.

PR #132, opened 2018-02-14
Pass bucket prefix back to form on error when creating bucket

andyhd:
* Added bucket prefix (`dev-` or `alpha-`) to template context on form error, because it was missing and resubmitting the form after an error would not add the prefix

## How to review

1. Create a bucket with the name of an existing bucket
2. See the form redisplayed with an error message
3. See that the prefix is still present

### DIFF
--- a/test/buckets/test-create.js
+++ b/test/buckets/test-create.js
@@ -28,4 +28,33 @@ describe('buckets/create', () => {
         assert.equal(redirect_url, url_for('buckets.details', { id: created_bucket.id }));
       });
   });
+
+  afterEach(() => {
+    delete process.env.ENV;
+  });
+
+  it('show an error if a bucket already exists with specified name', () => {
+    const form_data = {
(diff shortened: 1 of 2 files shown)


test/buckets/test-create.js
```diff
@@ -28,4 +28,25 @@ describe('buckets/create', () => {
...
+
+    mock_api()
+      .post('/s3buckets/')
+      .reply(400, errors);
+
+    return dispatch(handlers.create_bucket, { body: form_data })
+      .then(({ redirect_url, template, context }) => {
+        assert.isUndefined(redirect_url);
+        assert.equal(template, 'buckets/new.html');
+        assert.deepEqual(context.error.error, errors);
```
xoen: May be worth asserting that context passed to template also includes the expected `bucket`/`bucket_prefix` as you mostly have the test already in place.

andyhd: Oops, forgot that

PR #137, opened 2018-02-27
Adds breadcrumb component to main template and list pages

axemonkey:
## What

Adds the breadcrumb component from Elements to the template we use for all current pages, and implements it on the three superuser list pages. It can be added incrementally to other pages.

## How to review

1. log in as superuser
2. see a "Home" breadcrumb on any of the superuser function list pages

### DIFF
--- a/app/templates/apps/list.html
+++ b/app/templates/apps/list.html
@@ -1,14 +1,7 @@
 {% extends "layouts/one-column.html" %}
 
 {% set page_title = "Apps" %}
-
-{% block breadcrumbs %}
-<div class="breadcrumbs">
-  <ol>
-    <li><a href="/">Home</a></li>
-  </ol>
-</div>
-{% endblock %}
+{% set breadcrumbs = [{text: 'Home', url: '/'}, {text: 'List all webapps'}] %}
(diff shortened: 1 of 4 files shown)


app/templates/apps/list.html
```diff
@@ -2,6 +2,14 @@
 
 {% set page_title = "Apps" %}
 
+{% block breadcrumbs %}
+<div class="breadcrumbs">
+  <ol>
+    <li><a href="/">Home</a></li>
+  </ol>
+</div>
+{% endblock %}
```
andyhd: Since this is the same in all three templates, you could put it in an include, or a macro

axemonkey: Macro, of course. Thx. 👍

app/templates/layouts/one-column.html
```diff
@@ -8,6 +8,8 @@
 
 <main id="content" role="main">
 
+  {% block breadcrumbs %}{% endblock %}
```
andyhd: Or it could just go in here, and then be overridden in inheriting templates where necessary

PR #140, opened 2018-03-13
Added healtcheck (/healthz) endpoint

xoen:
### Why
The helm chart currently uses the login endpoint as a surrogate healtcheck
endpoint but now it's not working anymore as we're using Auth0 hosted login
page + silent SSO.

### Ticket
https://trello.com/c/fPZGrSYt/695-1-add-healtcheck-to-cp-ui


xoen:
Cheers @andyhd

PR #142, opened 2018-03-14
New tab for data warehouse buckets

axemonkey:
## What

Create a new tab in the CP UI for "data warehouse" S3 buckets, and differentiate these from "webapp data" S3 buckets. The former tab to be visible to all users.

## How to review

1. Log in as a superuser
2. Create two new buckets
3. Use the API to make one of these new buckets a warehouse bucket (set `is_data_warehouse` to `true`)
4. Refresh the UI homepage and see that the warehouse bucket is on the warehouse buckets tab and the other is on the webapp data tab
5. Use the API to downgrade your user to a non-superuser and log in again
6. See that you are able to view the "Warehouse data" tab, and that the warehouse bucket you created is visible there

### DIFF
--- a/app/base/handlers.js
+++ b/app/base/handlers.js
@@ -19,10 +19,17 @@ exports.home = (req, res, next) => {
   Promise.all([Deployment.list(), User.get(req.user.auth0_id)])
     .then(([tools, user]) => {
       ns.run(() => {
+        const is_warehouse = bucket => !!bucket.s3bucket.is_data_warehouse;
+        const warehouse_buckets = user.users3buckets.filter(bucket => is_warehouse(bucket));
+        const webapp_buckets = user.users3buckets.filter(bucket => !is_warehouse(bucket));
         res.render('base/home.html', {
           tools,
           rstudio_is_deploying,
           user,
+          buckets: {
+            warehouse: warehouse_buckets,
+            webapp: webapp_buckets,
+          },
         });
       });
     })
--- a/app/base/handlers.js
+++ b/app/base/handlers.js
@@ -19,17 +19,18 @@ exports.home = (req, res, next) => {
   Promise.all([Deployment.list(), User.get(req.user.auth0_id)])
     .then(([tools, user]) => {
       ns.run(() => {
-        const is_warehouse = bucket => !!bucket.s3bucket.is_data_warehouse;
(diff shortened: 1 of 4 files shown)


app/base/handlers.js
```diff
@@ -19,10 +19,17 @@ exports.home = (req, res, next) => {
   Promise.all([Deployment.list(), User.get(req.user.auth0_id)])
     .then(([tools, user]) => {
       ns.run(() => {
+        const is_warehouse = bucket => !!bucket.s3bucket.is_data_warehouse;
+        const warehouse_buckets = user.users3buckets.filter(bucket => is_warehouse(bucket));
+        const webapp_buckets = user.users3buckets.filter(bucket => !is_warehouse(bucket));
```
andyhd: You could sort the buckets into two groups like this:
```javascript
const buckets = user.users3buckets.reduce(function (buckets, bucket) {
  const group = !!bucket.s3bucket.is_data_warehouse ? 'warehouse' : 'webapp';
  buckets[group] = buckets[group] || [];
  buckets[group].push(bucket);
  return buckets;
}, {});

res.render(..., {
  ...,
  buckets,
});
```

axemonkey: ... also known as "Andy, could you tidy this up for me please?"

😀

Thanks. 👍

app/templates/buckets/list.html
```diff
@@ -1,31 +1,37 @@
...
-{% set page_title = "App data sources" %}
+{% set page_title = "Data sources" %}
 {% set breadcrumbs = [{text: 'Home', url: '/'}, {text: 'List all data sources'}] %}
 
 {% block main_column %}
 
-<p>{{ buckets.length }} app data source{%- if buckets.length != 1 -%}s{%- endif -%}</p>
+<p>{{ buckets.length }} data source{%- if buckets.length != 1 -%}s{%- endif -%}</p>
+<!-- <p>{{ warehouse_buckets.length }} warehouse data source{%- if warehouse_buckets.length != 1 -%}s{%- endif -%}</p>
+<p>{{ app_buckets.length }} app data source{%- if app_buckets.length != 1 -%}s{%- endif -%}</p> -->
```
andyhd: Commented out code should be removed. Also, when putting comments in templates, consider using template comment tags, eg: `{# comment #}`

axemonkey: 👍 Forgot. Ta.

PR #143, opened 2018-03-19
Create warehouse bucket

axemonkey:
## What

The API allows for data warehouse buckets now, and a tab exists to list them. This PR allows users (all users, not just superusers) to create them.

## How to review

1. Log in and see that you can access the "Warehouse data" tab
2. Create a new warehouse bucket
3. Check that it appears in the relevant tab
4. Log in as a superuser and check that your bucket does not appear on the "Webapp data" tab
5. Also check that your bucket does not appear in the bucket dropdown on the create app page

### DIFF
--- a/app/buckets/routes.js
+++ b/app/buckets/routes.js
@@ -3,7 +3,7 @@ const handlers = require('./handlers');
 
 module.exports = [
   { name: 'list', pattern: '/buckets', handler: handlers.list_buckets },
-  { name: 'new', pattern: '/buckets/new/:type', handler: handlers.new_bucket },
+  { name: 'new', pattern: '/buckets/new', handler: handlers.new_bucket },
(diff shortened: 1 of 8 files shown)


app/buckets/routes.js
```diff
@@ -3,7 +3,7 @@ const handlers = require('./handlers');
 
 module.exports = [
   { name: 'list', pattern: '/buckets', handler: handlers.list_buckets },
-  { name: 'new', pattern: '/buckets/new', handler: handlers.new_bucket },
+  { name: 'new', pattern: '/buckets/new/:type', handler: handlers.new_bucket },
```
andyhd: I think this URL should be either `/buckets/:type/new` or `/buckets/new?type=<type>` because REST: if you were listing buckets of a type, you would expect the URL to be `/buckets/:type/` or `/buckets?type=<type>`

axemonkey: 👍

PR #145, opened 2018-03-23
Data warehouse bucket page enhancements

axemonkey:
## What

Covers the first part of https://github.com/ministryofjustice/analytics-platform-control-panel/issues/108 - "Data warehouse bucket details page"

## How to review

Check that the changes described for the bucket details page are implemented and working.

1. Users in access group table (and in typeahead - see below) should show `name` and `username`
2. Users in access group table should display their access level - read only, read/write, or admin

screen shot 2018-03-27 at 16 15 26

The following apply when viewing a warehouse bucket to which you have admin access. (Note, if you are a superuser you will have automatic admin access to everything.)

1. Typeahead for adding users to data access group
image
2. You should be able to grant the three levels of access to users - users given 'admin' access will also be able to grant/revoke access to other users
image
3. You should be able to edit the access level for a given current user, or revoke access entirely
image

Note there are no protections against:
* removing your own access from a bucket
* removing the last user from a bucket
* removing the last admin from a bucket

### DIFF
--- a/app/buckets/handlers.js
+++ b/app/buckets/handlers.js
@@ -42,13 +42,10 @@ exports.create_bucket = (req, res) => {
 exports.bucket_details = (req, res, next) => {
   Promise.all([Bucket.get(req.params.id), App.list(), User.list()])
     .then(([bucket, apps, users]) => {
-      const current_user_is_bucket_admin = bucket.has_admin(req.user.auth0_id);
-
       res.render('buckets/details.html', {
         bucket,
         apps_options: apps.exclude(bucket.apps),
         users_options: users.exclude(bucket.users),
-        current_user_is_bucket_admin,
       });
     })
     .catch(next);
--- a/app/buckets/handlers.js
+++ b/app/buckets/handlers.js
@@ -42,10 +42,13 @@ exports.create_bucket = (req, res) => {
 exports.bucket_details = (req, res, next) => {
   Promise.all([Bucket.get(req.params.id), App.list(), User.list()])
     .then(([bucket, apps, users]) => {
+      const current_user_is_bucket_admin = bucket.has_admin(req.user.auth0_id);
+
       res.render('buckets/details.html', {
         bucket,
         apps_options: apps.exclude(bucket.apps),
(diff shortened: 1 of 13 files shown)


app/buckets/handlers.js
```diff
@@ -42,10 +42,13 @@ exports.create_bucket = (req, res) => {
 exports.bucket_details = (req, res, next) => {
   Promise.all([Bucket.get(req.params.id), App.list(), User.list()])
     .then(([bucket, apps, users]) => {
+      const current_user_is_bucket_admin = bucket.has_admin(req.user.auth0_id);
```
andyhd: Since `bucket` is already passed to the template, you could just call `bucket.has_admin()` there instead of passing a new context variable.
I would be tempted to add a `is_bucket_admin(bucket)` method to `User` and just call it in the template like `current_user.is_bucket_admin(bucket)`, so that the logic in the template is `{% if current_user.is_bucket_admin() or current_user.is_superuser %}`

app/templates/buckets/details.html
```diff
@@ -7,47 +7,78 @@
...
+
+                <button class="js-close-panel">Close panel</button>
+
+                <form action="{{ url_for('users3buckets.update', { id: users3bucket.id }) }}" method="post" class="inline-form clearfix">
+                  <input type="hidden" name="redirect_to" value="{{ url_for('buckets.details', { id: bucket.id }) }}" />
+                  <div class="form-group">
+                    <fieldset>
+                      <legend class="visuallyhidden">Data access level</legend>
+                      <div class="multiple-choice">
+                        <input type="radio" id="users3bucket_{{ users3bucket.id }}_data_access_level-readonly" name="users3bucket_{{ users3bucket.id }}_data_access_level" value="readonly" {% if users3bucket.access_level == 'readonly' %}checked{% endif %}>
```
andyhd: If these are all separate forms, you could remove the `{{ users3bucket.id }}` from the `name` attribute and simplify the handler code.

axemonkey: Not from the `id` though, because unique.

PR #151, opened 2018-04-12
Hide debug panels from non-superusers (on alpha)

axemonkey:
## What

Makes sure that the debug `<details>` panels are hidden from non-superusers on alpha in case they reveal more than they should (such as access/refresh tokens and the like).

The exception is on the `errors/internal-error.html` template, where it might be handy for users to be able to give us the information shown in the debug panel.

## How to review

1. Log in as a non-superuser
2. Check that debug panels are not visible except on the internal-error template

### DIFF
--- a/app/templates/apps/details.html
+++ b/app/templates/apps/details.html
@@ -153,7 +153,7 @@
 {% endif %}
 
 
-{% if current_user.is_superuser %}
+{% if current_user.is_superuser or env.ENV == 'dev' %}
   <details class="debug">
     <summary>Debug</summary>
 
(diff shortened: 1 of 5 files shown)


andyhd:
I think it would be more useful to only show these in dev, but for all users. Maybe show them in alpha for superusers too.

axemonkey:
OK yeah, good idea.

andyhd:
Actually, I've just thought, rather than having the `if` everywhere, it should be a template macro, eg:
```
{% debug %}
{{ some_vars }}
{% end debug %}
```
or something. But maybe that's for another time?

axemonkey:
Sure. 👍

PR #155, opened 2018-04-18
Add, list and delete customers

axemonkey:
## What

Trello: 
* [fix admin rights stuff that never got done](https://trello.com/c/2JSB6vv4/771-webapps-fix-the-admin-rights-stuff-that-never-got-done)
* [list app customers](https://trello.com/c/k8rfptae/755-list-customers-who-have-access-to-an-app)
* [app admins can add/remove customers](https://trello.com/c/xo0wMpIN/758-app-customers-admin-can-add-remove-customers)

## How to review

1. have an app in your control panel
2. add some customers
3. delete some customers

### DIFF
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -107,10 +107,8 @@ exports.details = (req, res, next) => {
   let users;
 
   Promise.all([App.get(req.params.id), Bucket.list(), User.list()])
-    .then(([returned_app, returned_buckets, returned_users]) => {
-      app = returned_app;
-      buckets = returned_buckets;
-      users = returned_users;
+    .then((results) => {
+      [app, buckets, users] = results;
       return app.customers;
     })
     .then((customers) => {
@@ -146,7 +144,7 @@ exports.delete_customer = (req, res, next) => {
       app.delete_customer(req.params.id, req.params.customer_id);
     })
     .then(() => {
-      res.redirect(req.body.redirect_to);
+      res.redirect(url_for('apps.details', { id: req.params.id }));
     })
     .catch(next);
 };
@@ -156,7 +154,7 @@ exports.add_customer = (req, res, next) => {
   App.get(req.params.id)
     .then(app => app.add_customer(req.params.id, req.body.customer_email))
     .then(() => {
-      res.redirect(req.body.redirect_to);
+      res.redirect(url_for('apps.details', { id: req.params.id }));
     })
     .catch((error) => {
       if (error.statusCode === 400 && error.error.email) {
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -104,13 +104,11 @@ exports.list = (req, res, next) => {
 exports.details = (req, res, next) => {
   Promise.all([App.get(req.params.id), Bucket.list(), User.list()])
     .then(([app, buckets, users]) => {
-      const current_user_is_app_admin = app.has_admin(req.user.auth0_id);
       res.render('apps/details.html', {
         app,
         buckets_options: buckets.exclude(app.buckets),
         users,
         users_options: users.exclude(app.users),
-        current_user_is_app_admin,
       });
     })
     .catch(next);
--- a/app/apps/handlers.js
+++ b/app/apps/handlers.js
@@ -137,3 +137,15 @@ exports.delete = (req, res, next) => {
     })
     .catch(next);
 };
+
+
+exports.delete_customer = (req, res, next) => {
+  App.get(req.params.id)
+    .then((app) => {
+      app.delete_customer(req.params.id, req.params.customer_id);
+    })
+    .then(() => {
+      res.redirect(req.body.redirect_to);
+    })
+    .catch(next);
+};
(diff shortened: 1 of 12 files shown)


app/apps/handlers.js
```diff
@@ -102,15 +102,25 @@ exports.list = (req, res, next) => {
...
+  let buckets;
+  let users;
+
   Promise.all([App.get(req.params.id), Bucket.list(), User.list()])
-    .then(([app, buckets, users]) => {
-      const current_user_is_app_admin = app.has_admin(req.user.auth0_id);
+    .then(([returned_app, returned_buckets, returned_users]) => {
+      app = returned_app;
+      buckets = returned_buckets;
+      users = returned_users;
```
andyhd: No need to change this, but it would be more concise (although perhaps not better) to do this like:
```javascript
.then((results) => {
  [app, buckets, users] = results;
```

app/models/control_panel_api.js
```diff
@@ -117,6 +117,19 @@ class App extends Model {
     return new ModelSet(User, this.data.userapps.map(ua => ua.user));
   }
 
+  get customers() {
+    const pk_name = this.constructor.pk;
+    const pk = this.data[pk_name];
```
andyhd: These can be combined into `const pk = this.data[this.constructor.pk];`

app/templates/apps/details.html
```diff
@@ -20,147 +15,230 @@ <h2 class="heading-medium">Repo</h2>
...
+</p>
+
+
+<div class="form-group">
+  <h2 class="heading-medium">App customers <em class="js-modal-trigger" data-content="app-customers"><span>What's this?</span></em></h2>
+  <div class="js-hidden js-modal-content" id="content-app-customers">
+    <h3 class="heading-small">App customers</h3>
+    <p>App customers are end users of a webapp. They do not have to be Control Panel users, any email address can be added.</p>
+  </div>
+  {% if customers.length %}
```
andyhd: Does this not work as `if customers`?

app/userapps/handlers.js
```diff
@@ -7,6 +7,10 @@ exports.create = (req, res, next) => {
 
   App.get(app_id)
     .then(app => app.grant_user_access(user_id, 'readonly', true))
+    .then(() => User.get(req.user.auth0_id))
+    .then((user) => {
+      req.session.passport.user.userapps = user.data.userapps;
+    })
```
andyhd: Since this has been repeated a few times, it would be worth extracting into a function:
```javascript
function update_current_user(req) {
  return User.get(req.user.auth0_id)
    .then((user) => {
      req.session.passport.user.userapps = user.data.userapps;
    });
}

// and then ...
App.get(app_id)
  .then(app => app.grant_user_access(...))
  .then(update_current_user(req))
  .then(() => {
    res.redirect(...);
  })
```

app/apps/handlers.js
axemonkey: Hmmm. I like it better though. I should try to remember about deconstruction more.

app/apps/handlers.js
```diff
@@ -128,3 +138,31 @@ exports.delete = (req, res, next) => {
...
 };
+
+
+exports.delete_customer = (req, res, next) => {
+  App.get(req.params.id)
+    .then((app) => {
+      app.delete_customer(req.params.id, req.params.customer_id);
+    })
+    .then(() => {
+      res.redirect(req.body.redirect_to);
```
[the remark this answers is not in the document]
axemonkey: No, I don't think so. Should we just redirect straight to the url_for the app page?

app/apps/routes.js
```diff
@@ -7,4 +7,6 @@ module.exports = [
   { name: 'create', method: 'POST', pattern: '/apps/create', handler: handlers.create },
   { name: 'details', pattern: '/apps/:id', handler: handlers.details },
   { name: 'delete', method: 'POST', pattern: '/apps/:id/delete', handler: handlers.delete },
+  { name: 'delete_customer', method: 'POST', pattern: '/apps/:id/customers/:customer_id/delete', handler: handlers.delete_customer },
+  { name: 'add_customer', method: 'POST', pattern: '/apps/:id/customers/add', handler: handlers.add_customer },
```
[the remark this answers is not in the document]
axemonkey: 👍

app/apps/handlers.js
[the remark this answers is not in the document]
andyhd: If it never changes, then I would be inclined to hard-code it here. Otherwise I think we'll need some defensive code to avoid a potential open redirect. I don't think it's actually a vulnerability, because the user has to be signed in to get here, but it's good practice.

app/templates/apps/details.html
andyhd: Oh, even in nunjucks?

andyhd: Yep, they diverge from Jinja2 on this - ok, never mind.

axemonkey: 👍

PR #156, opened 2018-04-23
Deep link to AWS buckets

axemonkey:
## What

Deep link to buckets on AWS using Kerin's AWS application (which is as yet not on alpha, but will be very soon).

## How to review

1. Create a bucket, either warehouse or webapp
2. See that there is a link to view it on AWS on the bucket list page
3. ...and the bucket detail page
4. ...and that they link through to AWS

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -43,9 +43,10 @@ config.auth0 = {
 };
 
 config.aws = {
-  login_url: {
-    prefix: `https://aws.services.${process.env.ENV}.mojanalytics.xyz/?destination=/s3/buckets/`,
-    suffix: '/?region=eu-west-1&tab=overview',
+  login_base_url: process.env.AWS_LOGIN_BASE_URL || `https://aws.services.${process.env.ENV}.mojanalytics.xyz`,
+  bucket_url: (name, region = 'eu-west-1') => {
+    const destination = escape(`/s3/buckets/${name}/?region=${region}&tab=overview`);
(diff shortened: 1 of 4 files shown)


app/config.js
```diff
@@ -43,7 +43,10 @@ config.auth0 = {
 };
 
 config.aws = {
-  login_url: process.env.AWS_LOGIN_URL,
+  login_url: {
+    prefix: `https://aws.services.${process.env.ENV}.mojanalytics.xyz/?destination=/s3/buckets/`,
+    suffix: '/?region=eu-west-1&tab=overview',
```
andyhd: Shouldn't the `destination` value be `escape`d? and the `region` and `tab` doubly escaped?

andyhd: Also, it would be better not to hard-code this base URL - the AWS login proxy thing might be deployed somewhere else (like on some `digital.justice.gov.uk` subdomain, for a likely example).
Better to split this up, like:
```javascript
config.aws = {
  login_base_url: process.env.AWS_LOGIN_BASE_URL || `https://aws.services.${process.env.ENV}.mojanalytics.xyz`,
  bucket_url: (name, region = 'eu-west-1') => {
    return `${config.aws.login_base_url}/?destination={escape(`/s3/buckets/${name}/?region=${region}&tab=overview`)}`,
  },
};
```
(might not be the best place for this function, but you get the idea

PR #161, opened 2018-05-08
Automatically refresh tools status (polling)

xoen:
### What

When the user is on the Tools tab, this panel content is updated every
few seconds (10 seconds) with the response from the `GET /tools` endpoint.

This `GET /tools` endpoint renders only the panel content (same template
used by the `home` template) but it's lighter as it only requests the tools
status.

**NOTE**: The polling is enabled only when the user selected the "Tools" tab and disabled when it's on other tabs so we don't make unnecessary requests.

### Why
Some users were expecting the tools status to refresh automatically and were confused by the fact the status didn't change.

### Technical details
The logic is in the `status-polling` module. The `tabs` module now emit a `tab-selected` event when a tab is selected, this event also contains the slug and the panel selected.

The `status-polling` module handle these events and enable/disable the polling accordingly (using `(set|clear)Interval`). Every 10 seconds the page makes an AJAX request to `GET /tools` and update the panel content with the new content from the response of this request.

### How to review

1. Go to the "Tools" tab
2. Deploy/Restart your RStudio
3. Wait ~10s and see the status change
4. Profit!

### Ticket

https://trello.com/c/2A3FU0uJ/787-4-polling-tools-status-in-cp

### DIFF
--- a/app/assets/javascripts/modules/status-polling.js
+++ b/app/assets/javascripts/modules/status-polling.js
@@ -1,0 +1,30 @@
+moj.Modules.toolsStatusPolling = {
+  toolsSlug: 'Analytical%20tools',
+  pollingFrequency: 10 * 1000,
+  pollingTimer: null,
+
+  init() {
+    this.bindEvents();
+  },
+
+  bindEvents() {
+    moj.Events.on('tab-selected', (event, params) => {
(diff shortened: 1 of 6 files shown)


xoen:
Brilliant, thanks @axemonkey 😄

PR #199, opened 2018-05-11
Grafana link

axemonkey:
## What

Simply puts a link (with `target="_blank"` so it opens in a new window) to Grafana on the tools page. The user's username is embedded in the link and Grafana has been tweaked to just show the time series (RAM and CPU) for that user.

## How to review

1. Check out and see the link on your tools page
2. Give it a gentle click, do not frighten it
3. See your CPU/RAM usage on Grafana in time series form and also in comparative bar graph form for context

Note, we only have Grafana on ALPHA env, so other envs show alternate text.

### DIFF
--- a/app/templates/tools/includes/list.html
+++ b/app/templates/tools/includes/list.html
@@ -58,7 +58,11 @@
   </table>
 
   <p>
-    You can <a href="https://grafana.services.alpha.mojanalytics.xyz/d/000000002/platform-users?refresh=10s&amp;orgId=1&amp;var-Username={{ current_user.username }}" target="_blank">view your resource utilisation on Grafana (opens in new tab)</a>.
+    {% if env.ENV == 'alpha' %}
+      You can <a href="{{ grafana_url }}" target="_blank">view your resource utilisation on Grafana (opens in new tab)</a>.
+    {% else %}
+      <em>(Grafana not available on environment: {{ env.ENV }})</em>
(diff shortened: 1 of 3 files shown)


app/templates/tools/includes/list.html
```diff
@@ -57,4 +57,8 @@ <h2 class="heading-medium">Your tools</h2>
     </tbody>
   </table>
 
+  <p>
+    You can <a href="https://grafana.services.alpha.mojanalytics.xyz/d/000000002/platform-users?refresh=10s&amp;orgId=1&amp;var-Username={{ current_user.username }}" target="_blank">view your resource utilisation on Grafana (opens in new tab)</a>.
```
xoen: As far as I can tell it looks good to me. I'd probably replace the hardcoded environment at least with the environment name from the config.

(You may want to add a separate config for the whole grafana URL but that's probably overkill for now as you can just derive it from what you already have)

xoen: You think it's worth urlencode the username? Probably it should be fine.

axemonkey: We don't have an equivalent `platform-users` dashboard in Grafana on dev, so the link wouldn't work.

xoen: I see, I think would be a bit cleaner maybe if the URL is in the config, but up to you really,

PR #208, opened 2018-05-18
Redirect users to /login when "state mismatch" errors occur

xoen:
### What
Users sometimes see these errors. We're now handling them more gracefully by:
* Clearing their session
* Clearing their session cookie and 
* redirecting the user to `/login`

(they'll need to login again if they're not logged in the "central" SSO)

### How to review

1. Start the app
2. Log in
3. Go to [http://localhost:3000/callback?code=BROKEN&state=BROKEN](http://localhost:3000/callback?code=BROKEN&state=BROKEN)
4. Check the logs for `info error-handler: Handling "state mismatch" error by redirecting user to login page...`
5. You should be redirected back to `/login`/logged in

### Ticket
https://trello.com/c/5czelJkJ/797-fix-internal-error-state-mismatch-error

### About `state`
[Auth0 documentation on `state` param](https://auth0.com/docs/protocols/oauth2/oauth-state)

### Also
I also removed the (direct) dependency to `uuid` as we don't need to generate the `state` value explicitly.

### DIFF
--- a/app/middleware/errors.js
+++ b/app/middleware/errors.js
@@ -2,6 +2,15 @@ const sentry = require('raven');
 const errlog = require('bole')('error-handler');
 
 
+const handle_state_mismatch = (req, res, conf) => {
+  errlog.info('Handling "state mismatch" error by redirecting user to login page...');
+
+  req.session.destroy(() => {
+    res.clearCookie(conf.session.name);
+    res.redirect('/login');
+  });
+};
+
 module.exports = (app, conf, log) => {
   log.info('adding errors');
 
@@ -12,6 +21,12 @@ module.exports = (app, conf, log) => {
 
   return (err, req, res, next) => {
     errlog.error(err);
+
+    if (err.message.startsWith('state mismatch')) {
+      handle_state_mismatch(req, res, conf);
+      return true;
+    }
+
     sentry.captureException(err);
 
     if (res.headersSent) {
--- a/app/middleware/errors.js
+++ b/app/middleware/errors.js
@@ -2,14 +2,14 @@ const sentry = require('raven');
 const errlog = require('bole')('error-handler');
 
 
-const handle_state_mismatch = (req, res, conf) => {
+function handle_state_mismatch(req, res, conf) {
   errlog.info('Handling "state mismatch" error by redirecting user to login page...');
 
   req.session.destroy(() => {
     res.clearCookie(conf.session.name);
     res.redirect('/login');
   });
-};
+}
 
 module.exports = (app, conf, log) => {
   log.info('adding errors');
--- a/test/middleware/test-errors.js
+++ b/test/middleware/test-errors.js
@@ -1,0 +1,43 @@
+const { assert } = require('chai');
+const app = require('express')();
+const config = require('../../app/config');
+const log = require('bole')('middleware');
+const errors_middleware = require('../../app/middleware/errors')(app, config, log);
+
+
+describe('When a "state mismatch" error occurs', () => {
+  it('redirects the user to /login', () => {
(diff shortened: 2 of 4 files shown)


app/middleware/errors.js
```diff
@@ -2,6 +2,15 @@ const sentry = require('raven');
 const errlog = require('bole')('error-handler');
 
 
+const handle_state_mismatch = (req, res, conf) => {
```
andyhd: This should be `function handle_state_mismatch(req, res, conf) {`

app/middleware/errors.js
```diff
@@ -12,6 +21,12 @@ module.exports = (app, conf, log) => {
 
   return (err, req, res, next) => {
     errlog.error(err);
+
+    if (err.message.startsWith('state mismatch')) {
```
andyhd: I think this `if` should be inside `handle_state_mismatch`, which should return true if there was a state mismatch that was handled.

test/middleware/test-errors.js
```diff
@@ -0,0 +1,43 @@
...
+
+
+describe('When a "state mismatch" error occurs', () => {
+  it('redirects the user to /login', () => {
+    let spy_session_destroyed = false;
+    let spy_cookie_cleared = 'NOT CLEARED';
+    let spy_redirected_to = 'NOT REDIRECTED';
+    let spy_next_called = false;
+
+    const err = Error('state mismatch, could not find [...]');
```
andyhd: Is this unfinished? Should there be a variable interpolation instead of `[...]`?

test/middleware/test-errors.js
```diff
@@ -0,0 +1,43 @@
...
+    };
+    const res = {
+      clearCookie: (name) => {
+        spy_cookie_cleared = name;
+      },
+      redirect: (url) => {
+        spy_redirected_to = url;
+      },
+    };
+    const next = () => {
```
andyhd: We already have a dependency on [`chai-spies`](https://github.com/chaijs/chai-spies), so you could use that to make this code simpler.

test/middleware/test-errors.js
xoen: The original error message is quite long and we don't really care in the context of testing, so I just cut the original message as the logic only checks if the error message starts with `state mismatch`.

test/middleware/test-errors.js
xoen: I'm not familiar with it, I'll give it a play to see if it makes things more complicated.

app/middleware/errors.js
xoen: I've seen both being used but I guess in this case there is no reason to stick the function in a constant really, I'll change 👍

app/middleware/errors.js
xoen: I'll play with it 👍

test/middleware/test-errors.js
andyhd: 👍

app/middleware/errors.js
andyhd: Here's a good couple of reasons to prefer `function`, at least at the top level of the module: https://medium.freecodecamp.org/constant-confusion-why-i-still-use-javascript-function-statements-984ece0b72fd

app/middleware/errors.js
xoen: I played with it a little bit but it made the code a bit more complicated so I'll leave it as it is as discussed offline.

xoen:
FYI: Learn how to use [chai-spies](https://github.com/ministryofjustice/analytics-platform-control-panel-frontend/pull/208/commits/137d7a312a15d41dc3122dd3c969616933f110b9) today 😄

PR #213, opened 2018-05-22
Re-enable docker image build and push to Quay.io

xoen:
### What
It was [~accidentally~ disabled here (see reason below)](https://github.com/ministryofjustice/analytics-platform-control-panel-frontend/pull/172/files) so images for new commits are not built/pushed to Quay.io.

I mostly copied from that commit with a couple of exceptions:
- We still don't want to push to Jenkins (which is now dead)
- Hence, we don't need to install curl, used to submit the Jenkins job
- Added the `IMAGE` environment variable to improve readability and reduce
  duplication
- Renamed step `deploy` to `docker_build_and_push`
- Use `run` step instead of `deploy` to better express intent and avoid
  confusion
- Renamed workflow from `test` to `test-build-and-push-docker-image` to
  be more explicit

### Why Docker build/push was disabled
The reason why @andyhd disabled the docker build/push was because we're replacing Jenkins with Concourse. The Concourse pipeline will be responsible for the Docker build/push to Quay.io.

### How to review

1. Check the CircleCI logs
2. Do you spot anything wrong?
3. Is the new Docker image available in [quay.io](https://quay.io/repository/mojanalytics/control-panel-frontend?tab=tags)

### Ticket
This problem just popped out, should I create a ticket?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,8 +16,53 @@ jobs:
             - node_modules
             - static
 
+  docker_build_and_push:
+    docker:
+      - image: docker:17.06.0-ce-git
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - setup_remote_docker
+      - restore_cache:
+          keys:
+            - image-cache-{{ .Branch }}
+            - image-cache-
+          paths:
+            - /caches/app.tar
+      - run:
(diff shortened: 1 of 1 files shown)


xoen:
Good idea @andyhd - the tag is somewhere there but output is not very clear so I'll add 👍

PR #221, opened 2018-05-31
wip

axemonkey:
Title the PR to complete the sentence: "Merging this PR will ..."

## What

Describe what you have changed and *why*

## How to review

1. Describe the steps required to test the changes
2. ???
3. Profit!

Provide [http://example.com](links) to relevant tickets, articles or other
resources.

Note the Trello Github integration links tickets to PRs/branches/commits (in
both directions).


app/assets/javascripts/modules/get-access-logs.js
```diff
@@ -0,0 +1,73 @@
+moj.Modules.accessLogs = {
+  dropdownName: 'js-access-log-period-dropdown',
+
+  init() {
+    this.$dropdown = $(`#${this.dropdownName}`);
```
andyhd: Since this is only used once, you could have `dropdownId: '#js-access-log-...'` and `this.$dropdown = $(this.dropdownId);`

app/assets/javascripts/modules/get-access-logs.js
```diff
@@ -0,0 +1,73 @@
...
+    if (this.$dropdown.length) {
+      this.$form = this.$dropdown.closest('form');
+      this.bindEvents();
+      this.getAccessLogs();
+    }
+  },
+
+  bindEvents() {
+    this.$dropdown.on('change', () => {
+      this.getAccessLogs();
```
andyhd: Can this be rewritten as `this.$dropdown.on('change', this.getAccessLogs);`?

app/assets/javascripts/modules/get-access-logs.js
```diff
@@ -0,0 +1,73 @@
...
+    });
+  },
+
+  getAccessLogs() {
+    const days = this.$dropdown.val();
+    const url = this.$form.attr('action');
+    const data = {};
+
+    if (parseInt(days, 10)) {
+      data.num_days = parseInt(days, 10);
```
andyhd: Doing the calculation twice. Could be cleaner. Maybe
```javascript
const data = {'num_days': parseInt(days, 10)};
if (!data.num_days) {
  data.num_days = default_days;
}
```

app/buckets/handlers.js
```diff
@@ -49,25 +49,27 @@ exports.create_bucket = (req, res) => {
...
-  let bucket;
-  let users;
   Promise.all([Bucket.get(req.params.id), User.list()]) // need to include App.list() in future
-    .then(([returned_bucket, returned_users]) => {
-      bucket = returned_bucket;
-      users = returned_users;
-    })
-    .then(() => bucket.access_logs)
-    .then((access_logs) => { // need to include apps in future
+    .then(([bucket, users]) => { // need to include apps in future
```
andyhd: Is this comment necessary/true/relevant any more?

app/buckets/handlers.js
```diff
@@ -49,25 +49,27 @@ exports.create_bucket = (req, res) => {
...
     })
     .catch(next);
 };
 
 
+exports.access_logs = (req, res, next) => {
+  Bucket.get(req.params.id)
+    .then(bucket => bucket.access_logs(req.query.num_days))
+    .then((access_log_data) => {
+      res.send(access_log_data);
```
andyhd: Should we send a `Content-Type: application/json` header too?

app/buckets/routes.js
```diff
@@ -7,5 +7,6 @@ module.exports = [
   { name: 'create', pattern: '/buckets/new', handler: handlers.create_bucket, method: 'POST' },
   { name: 'details', pattern: '/buckets/:id', handler: handlers.bucket_details },
   { name: 'delete', method: 'POST', pattern: '/buckets/:id/delete', handler: handlers.delete },
+  { name: 'access_logs', pattern: '/buckets/:id/access_logs/', handler: handlers.access_logs },
```
andyhd: Not sure if the trailing slash on the pattern is a problem, but none of the other patterns have one, so I'd say remove it for consistency.

app/assets/javascripts/modules/get-access-logs.js
axemonkey: 👍

app/buckets/handlers.js
```diff
@@ -49,25 +49,27 @@ exports.create_bucket = (req, res) => {
...
-        access_logs,
       }); // need to include apps_options: apps.exclude(bucket.apps) in future
     })
     .catch(next);
 };
 
 
+exports.access_logs = (req, res, next) => {
+  Bucket.get(req.params.id)
+    .then(bucket => bucket.access_logs(req.query.num_days))
```
andyhd: We should probably restrict `req.query.num_days` to one of the drop-down values, to prevent a user passing an arbitrarily high number for some kind of DoS attack. (The API should do this to, but it's good practice).

app/assets/javascripts/modules/get-access-logs.js
axemonkey: Seemingly no – trying that, when `getAccessLogs` runs, the reference to `this` is then referring to the HTML `<select>` element and not the parent module.

andyhd: Ah, I keep getting that wrong.

app/buckets/handlers.js
axemonkey: Should we?

app/buckets/handlers.js
axemonkey: It's still true and relevant. When webapps gets opened up to non-superusers this will need to include `apps` and a filtered version called `apps_options`.

Whether it's necessary is debatable, but the reason for having it hasn't changed or gone away.

app/buckets/routes.js
axemonkey: 👍

app/buckets/handlers.js
andyhd: Ok, then

app/buckets/handlers.js
andyhd: We should

axemonkey: 👍

PR #224, opened 2018-06-04
Include tests in Dockerfile

andyhd:
## What

* By using a [multi-stage Docker build](https://docs.docker.com/develop/develop-images/multistage-build/), we can build the image, run the tests inside it and then have the build fail if the tests do not pass. If the tests succeed, the image is built with the test code removed.
* This makes it much simpler to run the tests in a Concourse pipeline.

## How to review

1. `docker build -t frontend .`
2. See the test output
3. `docker run -it frontend sh`
4. `ls && exit` to see that the tests are not in the container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8.11.2-alpine
+FROM node:8.11.2-alpine AS image
 
 MAINTAINER Andy Driver <andy.driver@digital.justice.gov.uk>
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -20,3 +20,11 @@ ENV NODE_RESTART "0"
...
 
 CMD ["/usr/local/bin/yarn", "start"]
+
+
+FROM image AS test
+ADD test test/
+RUN yarn run test
+
+
+FROM image
```
xoen: What is this last `FROM image` doing? Is it the last `FROM` that wins? so the final docker image will be same as `image`, e.g without the `test` stage.

xoen: Yep. From this [article on multi-stage docker build](https://blog.alexellis.io/mutli-stage-docker-builds/):

> The general syntax involves adding FROM additional times within your Dockerfile - whichever is the last FROM statement is the final base image.

Nice 👍

PR #226, opened 2018-06-07
Force page size zero

axemonkey:
## What

Adds `?page_size=0` to all API requests to `list` anything in order to work around the "100 users only" bug. Setting `page_size` to 0 means return all records.

## How to review

1. You're unlikely to have more than 100 of anything in local dev, but you can create for example 10 users, then
2. edit L47 of `models/control_panel_api.js` to say `page_size: 3`
3. then list all users, and you should only see 3
4. change it back to `page_size: 0` and you should see all your users

Note: This should be replaced in future with proper pagination, using a non-zero `page_size` and writing out pagination buttons.

### DIFF
--- a/app/models/control_panel_api.js
+++ b/app/models/control_panel_api.js
@@ -44,7 +44,7 @@ class Model extends base.Model {
   }
 
   static list() {
-    return this.cpanel.get(this.endpoint)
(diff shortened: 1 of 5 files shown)


xoen:
Late but looks good to me 👍

PR #229, opened 2018-06-12
Rstudio deployment obscured by another deployment

djsd123:
See: [Trello Issue](https://trello.com/c/kL1TJGu6)

## What

Edge case:
When a k8s deployment is present in a given namespace (regardless of what app it is) the
logic on the list template's `if` expression will always evaluate to __true__
and never go into the `else` block where we've placed the controls to
manage __rstudio__.
So for example if there is a k8s deployment present that is **not** __rstudio__ and an __rstudio__ k8s deployment has not been created yet?
The user has no way of launcing/deploying __rstudio__.

Workaround:

- Change logic by remove `else` block ending the `if`

- Set a bool variable to gate __Rstudio__ controls

- Loop through __tools__ list and check if __Rstudio__ is present

- `if` not present display __Rstudio__ Deploy context control

The map shows which tools are managed by the Control-Panel (Just __rstudio__ at the moment)

#### Before my change

1. Use the `docker-compose` environment as instructed [here](https://github.com/ministryofjustice/analytics-platform-control-panel-frontend/blob/master/README.md)
2. Ensure you do not have a k8s __rstudio__ deployment present in your workspace
- `helm ls <YOUR GITHUBUSERNAME>`
- `helm del --purge <YOUR GITHUBUSERNAME>-rstudio`
3. If not deployed already? Deploy the Jupyter chart into your namespace
- `helm install analytics-platform-helm-charts/charts/jupyter-lab --name jupyter-lab-<YOUR GITHUBUSERNAME> --set Username=<YOUR GITHUBUSERNAME> --set aws.iamRole=alpha_user_<YOUR GITHUBUSERNAME> --namespace=user-<YOUR GITHUBUSERNAME> -f analytics-platform-config/chart-env-config/dev/jupyter.yml`
4. Browse to the Control-Panel Frontend. There you should see something similar to:
ScreenShot

#### After my change

Same as above except you should be able to **Deploy** __rstudio__

### DIFF
--- a/app/templates/tools/includes/list.html
+++ b/app/templates/tools/includes/list.html
@@ -37,7 +37,17 @@
         </td>
       </tr>
       {% endfor %}
-    {% else %}
+    {% endif %}
+
+    {% set showRstudio = true %}
+    {% for tool in tools %}
+      {% if tool.name == 'rstudio' %}
+        {% set showRstudio = false %}
+      {% endif %}
+    {% endfor %}
+    {% if showRstudio %}
+
+    <tr>
       <td>RStudio</td>
       <td>
         {% if rstudio_is_deploying %}
@@ -53,6 +63,8 @@
           </form>
         {% endif %}
       </td>
+    </tr>
+
     {% endif %}
     </tbody>
   </table>


app/base/handlers.js
```diff
@@ -29,6 +29,9 @@ exports.home = (req, res, next) => {
 
         res.render('base/home.html', {
           tools,
+          availableTools: [
+            { tool: 'rstudio', label: 'Rstudio' },
```
xoen: Not a biggie obviously but correct capitalisation is [RStudio](https://www.rstudio.com), missed my annoying comments :P?

app/templates/tools/includes/list.html
```diff
@@ -37,7 +37,7 @@ <h2 class="heading-medium">Your tools</h2>
         </td>
       </tr>
       {% endfor %}
-    {% else %}
+
```
xoen: If you remove the `if`/`else`, doesn't always loop through the tools and then shows the RStudio row (potentially twice)?

Maybe I'm missing something.

app/templates/tools/includes/list.html
```diff
@@ -46,14 +46,18 @@ <h2 class="heading-medium">Your tools</h2>
           Not deployed
         {% endif %}
       </td>
+        {% for availableTool in availableTools %}
+      <tr>
       <td class="align-right no-wrap">
         {% if not rstudio_is_deploying %}
```
xoen: Maybe this `if` should also check if tool currently rendered is RStudio, e.g.:

```python
{% if availableTool.tool == 'rstudio' and not rstudio_is_deploying %}
...
```

(or something like that)

app/templates/tools/includes/list.html
```diff
@@ -46,14 +46,18 @@ <h2 class="heading-medium">Your tools</h2>
...
-          <form action="{{ url_for('tools.deploy', { name: 'rstudio' }) }}" method="post" class="inline-form clearfix">
-            <input type="submit" class="button button-secondary right" value="Deploy" />
+          <form action="{{ url_for('tools.deploy', { name: availableTool.tool }) }}" method="post" class="inline-form clearfix">
+            <input type="submit" class="button button-secondary right" value="Deploy {{ availableTool.label }}" />
```
xoen: Given the name of the tool in that row is already in the first column, do we need to add it to the button label? It's a genuine question, it may be nice to safe some possibly precious space.

app/templates/tools/includes/list.html
```diff
@@ -37,7 +37,17 @@ <h2 class="heading-medium">Your tools</h2>
         </td>
       </tr>
       {% endfor %}
-    {% else %}
+    {% endif %}
+
+    {% set showRstudio = true %}
+    {% for tool in tools %}
+      {% if tool.name == 'rstudio' %}
```
xoen: As discussed offline, this probably needs to be `tool.app_label` but the rest looks good to me 👍

xoen: Oh, and I also think you could avoid this loop and use the one above (but don't worry about that!)

PR #244, opened 2018-06-28
Upgrade eslint-airbnb plugin and fix stuff accordingly

axemonkey:
## What

When upgrading some ESLint related dependencies, much stuff appeared to break. On further investigation it appears that the AirBNB ruleset we use has either added a rule or made one more strict, to do with the order of imported packages/objects. These have been fixed accordingly.

Mildly annoyingly, the latest version of `eslint-config-airbnb-base` has a peer dependency on an older version of `eslint` and so throws a warning. It's just a warning though, and everything works as it should.

### DIFF
--- a/app/api_clients/base.js
+++ b/app/api_clients/base.js
@@ -58,9 +58,7 @@ class APIClient {
       };
       log.debug(`${method} ${options.uri}`);
       return request(options)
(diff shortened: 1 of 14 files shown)


r4vi:
LGTM

PR #246, opened 2018-06-29
Upgrade github package to octokit/rest

r4vi:
Upgrade github package to octokit/rest

This couldn't happen automatically because of the way we were constructing the
proxy object for the github API.

The old version of the API returned an object that took a config and could be
`extended`, the new version of the api dynamically constructs the rest client by
going through it's own list of plugins and attaching methods to itself. This
didn't play well with the es6 extends syntax and after the version bump the
`authenticate` method we overrode was still pointing to the library version.

I've changed to shadowing it in the constructor instead of overriding it. This
seems to work.

resolves #178

https://trello.com/c/2jHQX3Gu/894-upgrade-github-dependency-to-octokit-rest

1. go to the apps/new endpoint
2. try to search for a github repo in the input box
3. the dropdown should have github repos

### DIFF
--- a/app/api_clients/github.js
+++ b/app/api_clients/github.js
@@ -1,4 +1,4 @@
-const GithubAPI = require('github');
+const GithubAPI = require('@octokit/rest');
 const { ManagementClient } = require('auth0');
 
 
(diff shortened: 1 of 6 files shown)


axemonkey:
Just needed to add the repo prefix hardcoded into the template, since it's no longer coming from config. 😄

axemonkey:
LGTM 👍

axemonkey:
BTW @r4vi , branch names with a slash in them always fail the `docker_build_and_push` task. It's ok to merge. 👍

PR #257, opened 2018-07-09
Fix parsing of Python tracebacks

andyhd:
## What

The code to nicely render Python tracebacks in dev was failing to parse `During handling of the above exception, another exception occurred` lines. This fixes that, which should make debugging easier.


app/api_clients/control_panel_api.js
```diff
@@ -42,31 +42,39 @@ class DjangoError extends APIError {
   constructor(error) {
     super(error);
     this.original_message = this.message;
+    const pattern = /500 - "(.*)Request Method:/;
+    const match = this.original_message.match(pattern);
+    this.message = match[1]
+      .trim()
+      .replace(/\\n/g, '<br>\n');
```
r4vi: this is quite hairy parsing, can we have a test for it with example of input?

app/api_clients/control_panel_api.js
```diff
@@ -42,31 +42,39 @@ class DjangoError extends APIError {
...
-      return { file, func, line_no, code };
-    });
-  }
+      match = line.match(pattern);
+      if (match) {
+        const [file, func, line_no, code] = match.slice(1);
+        return { file, func, line_no, code };
+      }
 
-  get message() {
```
xoen: This `message()` is not used anywhere right?

app/api_clients/control_panel_api.js
```diff
@@ -42,31 +42,39 @@ class DjangoError extends APIError {
...
   }
 
   static match(error) {
     return error.message && error.message.indexOf('Traceback:') >= 0;
   }
 
   get python_traceback() {
-    let pattern = /\\nTraceback:\s+\\n\\n(.*)\\n\\nException Type/;
-    const match = this.original_message.match(pattern);
+    if (!this.original_message) {
```
xoen: Isn't `this.original_message` always there? I may miss something.

andyhd: It could be `null` or `undefined`

app/api_clients/control_panel_api.js
andyhd: `get message()` has been replaced by `this.message` on line 47

app/api_clients/control_panel_api.js
andyhd: Yeah, I'll write a test.

PR #292, opened 2018-10-10
Update grafana dashboard URL

kerin:
The Grafana dashboard URL has changed, because it generates an internal id as part of the URL by default - that has been changed to use a static value. See matching PR in [`analytics-platform-config`](https://github.com/ministryofjustice/analytics-platform-config/pull/92).

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -109,7 +109,7 @@ config.express = {
 };
 
 config.grafana = {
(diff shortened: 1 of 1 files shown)


kerin:
Added `donotmerge` label to indicate that this shouldn't be merged or deployed until the Prometheus/Grafana changes are ready to go.

PR #294, opened 2018-11-02
Fixes links to AWS and Grafana, when on platform domains other than mojanalytics.xyz

davidread:
## What

The URL for these links use "SERVICES_DOMAIN" variable instead of hardcoded one. I'm doing this for accelerator.

## How to review

I will deploy to dev and try the grafana and S3.


davidread:
old

PR #308, opened 2019-01-31
Add webapp link on details page

andyhd:
## What

* Add a link to the webapp deployment at the top of the app details page


davidread:
Yes, great! My preference would be "Logs" rather than "App logs" (and arguably all the others that are "App X" too)
screen shot 2019-02-04 at 09 53 51

PR #310, opened 2019-02-06
Let superusers edit users and reset MFA

andyhd:
## What

* Allow superusers to toggle a user's superuser status
* Allow superusers to reset a user's MFA settings
* Allow superusers to delete a user


app/templates/users/edit.html
```diff
@@ -1,13 +1,57 @@
...
+    <div class="form-group">
+      <input type="submit" class="button button-secondary" value="Reset MFA">
+    </div>
+  {%- endcall %}
+
+</form>
+
+<hr>
+
+<form action="{{ url_for("users.delete", { id: user.auth0_id }) }}" method="post">
```
davidread: It would good to explain that this is useful as a tidy up only if the user is removed from the github org, otherwise they can simply log-in again and their user is recreated. Maybe we should also say whether this delete also deletes their app bucket permissions, AWS role and NFS home directory. This could all be a follow up ticket.

app/templates/users/details.html
```diff
@@ -12,8 +18,50 @@
...
+{% endif %}
 
 {% include 'users/includes/tables.html' %}
 
+{% if current_user.is_superuser %}
+<hr>
+
+<form action="{{ url_for("users.delete", { id: user.auth0_id }) }}" method="post">
+
+  {{ delete_button("Delete user", confirm="Are you sure you want to delete this user?") }}
```
xoen: I can't remember how we handle this in the backend (CP-API), what are the implications of deleting a user? Is it just "delete from DB" or we do more, e.g. uninstall tools helm charts, uninstall init/config helmcharts, namespace, etc...?

app/templates/users/edit.html
```diff
@@ -1,13 +1,57 @@
...
+    </div>
+  {%- endcall %}
+
+  <div class="form-group">
+    <input type="submit" class="button button-secondary" value="Save changes">
+  </div>
+
+</form>
+
+<form method="{{ url_for('users.reset_mfa', { id: user.auth0_id | urlencode }) }}" method="post">
```
xoen: Maybe there is some room for factoring these forms in a template so we don't have to repeat them in different views/templates.

test/users/test-edit.js
```diff
@@ -1,33 +0,0 @@
-const { assert } = require('chai');
```
davidread: surely we want more tests for the new options, not fewer?

app/users/handlers.js
```diff
@@ -27,13 +27,24 @@ exports.user_details = (req, res, next) => {
...
+
+      return user.update()
+        .then(() => {
+          req.session.flash_messages.push('Updated user');
+
+          let url = url_for('users.details', { id: user.auth0_id });
+
+          if (user.auth0_id === req.user.auth0_id) {
+            // force re-login
+            url = url_for('base.logout');
```
xoen: Why do we force re-login? I imagine the frontend app caches the user data somewhere so that would need updating but is there any other reason why users would need to re-login if they update themselves? Does it has anything to do with tokens?

app/users/routes.js
```diff
@@ -5,8 +5,9 @@ module.exports = [
   { name: 'new', pattern: '/users/new', handler: handlers.new_user },
   { name: 'list', pattern: '/users', handler: handlers.list_users },
   { name: 'details', pattern: '/users/:id', handler: handlers.user_details },
-  { name: 'edit', pattern: '/users/:id/edit', handler: handlers.user_edit },
+  { name: 'edit', method: 'POST', pattern: '/users/:id/edit', handler: handlers.user_edit },
```
xoen: I haven't seen this in AGES, so maybe I can't remember correctly, but wasn't the `GET users.edit` route for "I want to edit a user, show me the edit form"?

So maybe we need another route for the actual update where user submit the update data? e.g. `PUT users.update`?

app/templates/users/details.html
```diff
@@ -12,8 +18,50 @@
 
 {% block main_column %}
 
+{% if current_user.is_superuser %}
+<form action="{{ url_for('users.edit', { id: user.auth0_id | urlencode }) }}" method="post">
```
xoen: As I just mentioned in other comment, this probably will need to submit to a new route called `PUT users.update` instead of changing existing route to view edit form.

app/templates/users/details.html
andyhd: The delete user button already exists in the user list page. I just put a copy on the user details page as well. Changing what it means to delete a user is not in scope of this change and would be in the control panel api anyway.

app/templates/users/edit.html
andyhd: I'm not sure what you mean. I already moved most of the repeated code into some nunjucks macros.

app/users/handlers.js
```diff
@@ -27,13 +27,24 @@ exports.user_details = (req, res, next) => {
...
-        apps_options: apps.exclude(user.apps),
-        buckets_options: buckets.exclude(user.buckets),
-      });
+  User.get(req.params.id)
+    .then((user) => {
+      user.is_superuser = !!req.body.superadmin;
+
+      return user.update()
+        .then(() => {
+          req.session.flash_messages.push('Updated user');
```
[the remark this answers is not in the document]
andyhd: Yep

app/users/handlers.js
andyhd: If you remove your own superuser status, you are forced to re-login, as you are still a superuser in your current session. I could modify the user object in the session instead, and redirect to the home page, which would achieve the same sort of thing without having to login again, I suppose.

app/templates/users/edit.html
andyhd: Deleting users is outside the scope of this change - deleting bucket permissions, etc is a change for the control panel api, not the frontend. I just put the delete user page on the user details page as it seems like a sensible place for it - it already exists on the user list page.

test/users/test-edit.js
andyhd: This test didn't really test anything anyway. I'll add some new tests.

app/users/routes.js
andyhd: The forms for editing the user are on the `users.details` page. Those forms POST to the `users.edit` handler. There was no user edit page, only a placeholder which showed a JSON dump of the user object.

app/templates/users/details.html
andyhd: Usually only `GET` and `POST` methods are used by HTML forms, as other verbs may not be supported by browsers (or at least historically). If you look at the other routes in the frontend, there are no "edit pages" for other objects (eg apps) - the forms are displayed on the details page and submit to POST-only update handlers.

app/templates/users/edit.html
xoen: Hang on, shouldn't this be `action` instead of `method`?

xoen: Maybe I'm wrong, but I see duplicated stuff in `users/details.html` and in `users/edit.html`

app/users/handlers.js
xoen: Yeah, that's what I meant. Unless the token needs to change or any of the other fields of the user object is not coming from the response and it's invalidate by this request or something like that.

app/users/routes.js
xoen: I see. Well, maybe it was just a Rails convention but naming this route `update` instead of `edit` probably would avoid any confusion as user is being _updated_ by the request.

Also - sorry I'm not trying to annoy you with this comment - shouldn't the method be `PUT` as users are updating not creating a resource. Not the end of the world if doesn't change of course: Computers don't care :D

app/templates/users/details.html
xoen: Yeah I know, that's annoying. I didn't think about that, for some reason I thought we were using some king of fancy wrapper to generate the form code with an hidden support field (like `_method`) or something like that, sorry I'm tired, you're right 👍

app/templates/users/edit.html
andyhd: Yes, I thought I committed that fix already, sorry.

andyhd: Ah, I should delete `users/edit.html` - it's not used any more. 👍

app/users/routes.js
andyhd: The frontend isn't a REST API, but if it was I would agree that `PUT` would be the correct verb. `update` is a better name for the handler 👍

app/templates/users/edit.html
andyhd: Good catch, by the way - it took me about an hour to track that one down earlier 🤦‍♂️

xoen: eheh you know by now that I notice the weirdest things 😆 - but yeah, there is a law of physics that says that the amount of time required for a fix is inversely proportional to the number of characters involved 🕳

PR #329, opened 2019-02-18
Fix auth bug

andyhd:
## What

* There was an error sometimes when logging in to the control panel:
  ```did not find expected authorization request details in session, req.session["oidc:alpha-analytics-moj.eu.auth0.com"] is undefined```
  This seems to be caused by a race-condition in ExpressJS, which can send the redirection headers to the client before the session has been stored. Then the client can load the redirection URL, which fails to retrieve the session because it hasn't been stored yet. This PR explicitly saves the session and only redirects when the save function returns.

* Fixed a minor issue where the `render_debug` template macro failed to display the default items (current user and session) because macros do not have access to the calling template context by default.

* Added a Makefile, because I forget which commands to run when switching between projects.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,45 @@
+HOST=0.0.0.0
+PORT=3000
+PROJECT=control-panel-frontend
+
+-include .env
+export
+
+.PHONY: collectstatic dependencies help run test wait_for_db
+
+## dependencies: Install dependencies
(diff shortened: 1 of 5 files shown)


r4vi:
I don't love `ensure_session_saved`'s name but can't think of anything better

